PR #5, opened 2020-10-11
Fix order references

### LINKED ISSUE #1105: Sort the result of "Find References" and related features
Hi,

That's something that bugs me for a while now so I wanted to leave a trace.
When looking for references, implementations, etc. in the code the result list is not sorted.
Which makes it quite difficult to quickly identify the different files impacted, I don't have a test project to gave an example but that's the kind of result we will get:
```
directory/something/File1.php:23
directory/somethingElse/File2.php:89
directory/something/File1.php:14
```

When it will make more sense to have something like:
```
directory/something/File1.php:14
directory/something/File1.php:23
directory/somethingElse/File2.php:89
```

camilledejoye:
Order the result of the find references feature.
Fix https://github.com/phpactor/phpactor/issues/1105

Depends on https://github.com/phpactor/text-document/pull/7 in this current implementation.

### DIFF
--- a/lib/LanguageServerReferenceFinder/Handler/ReferencesHandler.php
+++ b/lib/LanguageServerReferenceFinder/Handler/ReferencesHandler.php
@@ -161,18 +161,7 @@ class ReferencesHandler implements Handler, CanRegisterCapabilities
      */
     private function toLocations(array $locations): array
     {
-        usort($locations, function (Location $first, Location $second) {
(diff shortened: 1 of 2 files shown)


camilledejoye:
I just forced push with no modification to trigger the CI.
Passing, and the dependent PR is already merged.

Maybe we can merge this one and https://github.com/phpactor/phpactor/pull/1109 at the same time to keep the same user experience.

PR #6, opened 2020-11-28
fix completion result isComplete

camilledejoye:
Update from an old PR on the previous repository: https://github.com/phpactor/language-server-extension/pull/8

Currently we always return that the completion list is incomplete.
The PR use the return of the completor to define it in order for a client to be able to avoid unnecessary requests.


dantleech:
:bowing_man:

PR #12, opened 2020-12-24
Fix WorkspaceIndex not reindexing when document changes.

BladeMF:


lib/LanguageServerWorseReflection/Workspace/WorkspaceIndex.php
```diff
@@ -37,41 +46,63 @@ public function documentForName(Name $name): ?TextDocument
...
+            $this->byName[$name] = $textDocument;
+        }
+        foreach($namesToRemove as $name){
+            unset($this->byName[$name]);
+        }
+
+        if(!empty($newNames))
+            $this->documentToNameMap[(string)$textDocument->uri()] = $newNames;
+        else
+            unset($this->documentToNameMap[(string)$textDocument->uri()]);
```
dantleech, line 80: cs fixer is failing, probably this is one of them, just run `./vendor/bin/php-cs-fixer fix` (or `composer integrate`)

dantleech:
I'll need to check to see the performance regression (if any) with this.

BladeMF:
I wouldn't expect to have any regression in performance. We are not `foreach`-ing the index now, so it should be faster, but using a bit more memory.

lib/LanguageServerWorseReflection/Workspace/WorkspaceIndex.php
```diff
@@ -37,41 +45,64 @@ public function documentForName(Name $name): ?TextDocument
...
+        foreach ($newNames as $name) {
+            $this->byName[$name] = $textDocument;
+        }
+        foreach ($namesToRemove as $name) {
+            unset($this->byName[$name]);
+        }
+
+        if (!empty($newNames)) {
+            $this->documentToNameMap[(string)$textDocument->uri()] = $newNames;
+        } else {
```
dantleech, line 79: minor but I would remove else in favor of an early return:

```
if ($something) {
    // do something
    return;
}

// do something else
```

dantleech:
Good job :+1:

lib/LanguageServerWorseReflection/Workspace/WorkspaceIndex.php
BladeMF: Sure, but I am not sure how to append this to the current PS - push to the master branch again?

dantleech: It's fine, it's merged now.

PR #19, opened 2021-01-05
Create class support

dantleech:
Adds support for Phpactor's class generator functionality.

- The existing functionality will create a class according to the composer configuration.
- Add diagnostic which is shown in an empty file
- Adds a code action for _each registered class variant_ when the file is empty.

recording1


tests/LanguageServerCodeTransform/Unit/LspCommand/CreateClassCommandTest.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        return SourceCode::fromStringAndPath(self::EXAMPLE_TEXT, self::EXAMPLE_PATH);
+    }
+}
+
+class TestFileToClass implements FileToClass
+{
+    public const TEST_CLASS_NAME = 'Foobar';
+
+    public function fileToClassCandidates(FilePath $filePath): ClassNameCandidates
```
dantleech, line 70: would be good to add test implementations to these respective libraries

PR #21, opened 2021-01-08
Rename LSP support

BladeMF:
I am creating this PR so we can discuss the code. It's not going to get any better if I wait longer :-)

The code works by using _Find all references_ to find everything that can be renamed.
1. **Rename type**. If type has the same name as the file, **rename the file**.
2. **Rename local variable**.
3. **Rename method**.

https://github.com/phpactor/phpactor/issues/1140


.gitignore
```diff
@@ -5,3 +5,4 @@
 /composer.lock
 /.php_cs.cache
 /.phpbench
+/.vscode
```
dantleech, line 8: this has now been added to all phpactor packages

composer.json
```diff
@@ -62,6 +62,8 @@
             "Phpactor\\Extension\\LanguageServer\\Tests\\": "tests/LanguageServer/",
             "Phpactor\\Extension\\LanguageServerCompletion\\Tests\\": "tests/LanguageServerCompletion/",
             "Phpactor\\Extension\\LanguageServerReferenceFinder\\Tests\\": "tests/LanguageServerReferenceFinder/",
+            "Phpactor\\Extension\\LanguageServerRename\\Tests\\": "tests/LanguageServerRename/",
+            "Phpactor\\Extension\\LanguageServerSelectionRange\\Tests\\": "tests/LanguageServerSelectionRange/",
```
dantleech, line 66: added in error?

lib/LanguageServerHover/Handler/HoverHandler.php
```diff
@@ -176,7 +176,7 @@ private function renderMember(SymbolContext $symbolContext): string
     private function renderFunction(SymbolContext $symbolContext): string
     {
         $name = $symbolContext->symbol()->name();
-        $function = $this->reflector->reflectFunction($name);
+        $function = $this->reflector->reflectFunction((string)$symbolContext->name());
```
dantleech: no need for a type cast. the WR API can accept either `Name` or `string`

lib/LanguageServerRename/LanguageServerRenameExtension.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function load(ContainerBuilder $container)
+    {
+        $container->register(RenameHandler::class, function (Container $container) {
+            return new RenameHandler(
+                $container->get(LanguageServerExtension::SERVICE_SESSION_WORKSPACE),
+                new Parser(),
+                new Renamer(
```
dantleech: Extract this to separate service

lib/LanguageServerRename/LanguageServerRenameExtension.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function load(ContainerBuilder $container)
+    {
+        $container->register(RenameHandler::class, function (Container $container) {
+            return new RenameHandler(
+                $container->get(LanguageServerExtension::SERVICE_SESSION_WORKSPACE),
+                new Parser(),
```
dantleech: use `$container->get('worse_reflection.tolerant_parser')` : this service should be exposed via. a constant `SERVICE_TOLERANT_PARSER` but it's not. This is better than `new Parser` as it's a caching version.

lib/LanguageServerRename/Model/NodeUtils.php
```diff
@@ -0,0 +1,120 @@
...
+            PositionConverter::intByteOffsetToPosition($token->getStartPosition(), $document),
+            PositionConverter::intByteOffsetToPosition($token->getEndPosition(), $document)
+        );
+    }
+
+    public function getNodeNameToken(Node $node, ?string $name = null): ?Token
+    {
+        if ($node instanceof MethodDeclaration) {
+            return $node->name;
+        } elseif ($node instanceof ClassDeclaration) {
```
dantleech: all the `else` here are redundant as we `return` can replace with multiple `if` statements.

lib/LanguageServerRename/Model/NodeUtils.php
```diff
@@ -0,0 +1,120 @@
...
+class NodeUtils
+{
+    public function getNodeNameText(Node $node, string $documentContent): ?string
+    {
+        $token = $this->getNodeNameToken($node);
+        if ($token === null) {
+            return null;
+        }
+        $text = (string)$token->getText($documentContent);
+        if(mb_substr($text, 0, 1) == '$')
```
dantleech: this is specific to properties probably, and this logic should be with the property logic.

lib/LanguageServerRename/Model/Renamer.php
```diff
@@ -0,0 +1,314 @@
...
+        if ($this->canRenameNode($node)) {
+            return $this->nodeUtils->getNodeNameRange($node);
+        }
+
+        return null;
+    }
+
+    public function rename(TextDocumentItem $textDocument, Position $position, string $newName): ?WorkspaceEdit
+    {
+        [$offset, $node] = $this->documentAndPositionToNodeAndOffset($textDocument, $position);
```
dantleech: this method doesn't need to do two things. you can extract the offset independently above this line.

lib/LanguageServerRename/Handler/RenameHandler.php
```diff
@@ -0,0 +1,70 @@
...
+        return [
+            PrepareRenameRequest::METHOD => 'prepareRename',
+            RenameRequest::METHOD => 'rename',
+        ];
+    }
+
+    public function rename(RenameParams $params): Promise
+    {
+        return \Amp\call(function () use ($params) {
+            return $this->renamer->rename($this->workspace->get($params->textDocument->uri), $params->position, $params->newName);
```
dantleech, line 48: the `Renamer` needs to return a `Promise`, in which case this would look like `return yield $this->renamer->rename(...)`

lib/LanguageServerRename/Model/Renamer.php
```diff
@@ -0,0 +1,314 @@
...
+                    'Reference find stopped, %s/%s references confirmed but took too long (%s/%s seconds).',
+                    count($locations),
+                    $count,
+                    number_format(microtime(true) - $start, 2),
+                    $this->timeoutSeconds
+                ));
+                return $this->locationsToWorkspaceEdit($locations, $oldName, $newName);
+            }
+
+            // I can't see how I can yield here.
```
dantleech: the public `rename` method needs to return a `Promise` and this method needs to be a [co-routine](https://amphp.org/amp/coroutines/). Basically you need to wrap it in `call`:

```
/**
 * @return Promise<WorkspaceEdit|null>
 */
private function doRename($1, $2): Promise
{
    return call(function () use ($1, $2) {
        // ...
        yield delay(0); // maybe nicer to use the function rather than the class.
    });
}
```

lib/LanguageServerRename/Model/Renamer.php
```diff
@@ -0,0 +1,314 @@
...
+    
+    private function renameNode(TextDocument $phpactorDocument, ByteOffset $offset, Node $node, string $oldName, string $newName): ?WorkspaceEdit
+    {
+        if (empty($oldName)) {
+            return null;
+        }
+        $locations = [];
+        try {
+            $potentialLocation = $this->definitionLocator->locateDefinition($phpactorDocument, $offset);
+            $locations[] = new Location($potentialLocation->uri(), $potentialLocation->offset());
```
dantleech: the finder should also find this location, why do we do it here? (probably not a good idea to use the definition locator here, which may do more than we want and shouldn't be coupled to this service)

lib/LanguageServerRename/Model/Renamer.php
```diff
@@ -0,0 +1,314 @@
...
+        $this->parser = $parser;
+        $this->definitionLocator = $definitionLocator;
+        $this->finder = $finder;
+        $this->clientApi = $clientApi;
+        $this->workspace = $workspace;
+        $this->renameVariable = $renameVariable;
+        $this->nodeUtils = $nodeUtils;
+    }
+
+    public function prepareRename(TextDocumentItem $textDocument, Position $position): ?Range
```
dantleech, line 95: public APIs should accept Phpactor documents, in this case the text document and range from the Phpactor `phpactor/text-document` library. This would enable us to use this service outside of the LS (CLI / RPC)

lib/LanguageServerRename/Model/Renamer.php
```diff
@@ -0,0 +1,314 @@
...
+        return $this->locationsToWorkspaceEdit($locations, $oldName, $newName);
+    }
+
+    private function documentAndPositionToNodeAndOffset(TextDocumentItem $textDocument, Position $position): array
+    {
+        $offset = PositionConverter::positionToByteOffset($position, $textDocument->text);
+        
+        $rootNode = $this->parser->parseSourceFile($textDocument->text);
+        $node = $rootNode->getDescendantNodeAtPosition($offset->toInt());
+        return [$offset, $node];
```
dantleech: This method can simply reutrn `Node` if we remove the offset responsiblity

lib/LanguageServerRename/Model/Renamer.php
```diff
@@ -0,0 +1,314 @@
...
+    {
+        $edits = [];
+        $rename = null;
+        $documentContent = $this->getDocumentText($documentUri);
+        $rootNode = $this->parser->parseSourceFile($documentContent);
+        foreach ($locations as $location) {
+            /** @var Location $location */
+            $node = $rootNode->getDescendantNodeAtPosition($location->offset()->toInt());
+            
+            if(($r = $this->renameFileIfNeeded($node, $documentUri, $oldName, $newName)) !== null){
```
dantleech: would be happy if we removed the file-rename functionality for now, this is a more complex topic I think. Let's concentrate on a smaller set of features.

lib/LanguageServerRename/Model/Renamer.php
```diff
@@ -0,0 +1,314 @@
...
+                    new Range(
+                        $position,
+                        new Position($position->line, $position->character + mb_strlen($oldName))
+                    ),
+                    $newName
+                );
+            }
+        }
+
+        return [ $edits, $rename ];
```
dantleech: if we remove renmae file then this becomes returns only TextEdits

dantleech (commented):
Great that you've achieved what you have so far.

In general I think this is looking good, I want to suggest a different architetcture though:

```
Model\Rename.php // main class
Model\Renamer.php // interface
Model\Renamer\VariableRenamer.php
Model\Renamer\ClassLikeRenamer.php
Model\Renamer\MethodRenamer.php
```

This could be implemented by the `Rename` service having a collection of `Rename` instances, resolving the appropriate renamer for the node:

```php
$this->resolveRenamer($node)->rename(...);
```

```php
private function resolveRenamer(Node $node): Renamer
{
    foreach ($this->renamers as $renamer) {
        if (!$renamer->supports($node)) {
            continue;
        }
        return $renamer;
    }

    throw new CouldNotFindRenamer(sprintf('Could not find renamer for node "%s"', get_class($node));
}
```

We can start with _only_ implementing `VariableRenamer`: this will allow us to merge and build a solid base for implementing the remaining functionality - some of which is not trivial (e.g. class renaming aka. class _moving_)

some other comments:

- The _public_ API for the `Rename` service should accept the Phpactor value objects (`Phpactor\...\TextDocument` etc) ...
- ... as the `Renamer` class may be split into it's own "package" or otherwise extracted so that it can be used by the CLI and the legacy RPC implementation _at a later date_.

Some of the coding-standard is not what I'd expect - not sure if the CS fixer is working or not.

BladeMF:
Thank you for your notes and ideas! I appreciate your time with this.

Just so I am sure I understand your idea correctly, let me try to explain how it _currently_ works and then we can get to me understanding _your_ idea.

### How it currently works
1. Use the current `DefinitionLocator` to locate the definition using the current offset
2. Use the current `ReferenceFinder` to locate all the references at the current offset
3. Collect all the locations and group them by document
4. For each document, for each location:
4.1. Check if file needs renaming (are we renaming a type with the same name as the file?)
4.2. Get the precise range for each edit
5. Create a `WorkspaceEdit` object.

### Your idea
While I agree that class moving should be somewhere in there, I have trouble seeing how we can use this architecture while using `DefinitionLocator`, `ReferenceFinder` because they currently implement the task of splitting between class, members variables and methods and variables. 

Can you please elaborate more on how the above logic will go into three classes as it does not cite classes or members or variables with the exception of when deciding whether to rename the file. Do you want to reimplement the same reference finding logic in `VariableRenamer`, `ClassRenamer` and `MethodRenamer` or am I misunderstanding you?

composer.json
BladeMF: Not really, otherwise "Fix namespace..." didn't work properly.

lib/LanguageServerRename/LanguageServerRenameExtension.php
BladeMF: What do you mean?

lib/LanguageServerRename/Model/Renamer.php
BladeMF: I agree, but the `ReferenceFinder` doesn't return the definition. I took this code from the "Find all references" hander.

lib/LanguageServerRename/LanguageServerRenameExtension.php
dantleech: This should be registered independently:

```
$conatiner->register(Rename::class, function (Container) {
    return new Renamer( ... );
});
```

composer.json
BladeMF: I needed that and composer install (supposedly to update the autoloader) for it to work. The other namespace I added I think you've missed because tests have weird namespace (note the missing starting letter of Selection):
```php
namespace Phpactor\Extension\LanguageServer\Tests\electionRange\Unit\Model;
```

lib/LanguageServerRename/Model/NodeUtils.php
BladeMF: No, it includes variables as well. Putting it here enforces it as a policy - we don't show the `$` in the rename box and we don't expect it in the new name.

lib/LanguageServerRename/Model/Renamer.php
BladeMF: This logic will not survive anyway, but I agree.

lib/LanguageServerRename/Handler/RenameHandler.php
BladeMF: I agree, let's see how it turns out.

lib/LanguageServerRename/Model/Renamer.php
BladeMF: Yep, for the next version.

lib/LanguageServerRename/Model/Renamer.php
BladeMF: What do you mean, 'more complex'? Specifically renaming when renaming the class itself is pretty straight-forward - either they have the same name or they don't (unless I am missing something). The other way around is the complicated bit, I think - when the class file is being moved..

lib/LanguageServerRename/Model/Renamer.php
BladeMF: That is going to be reworked completely (we'd be returning `RenameResult`s) so no need to clear that now, I think.

lib/LanguageServerRename/Model/NodeUtils.php
dantleech: Right, but this is rule only applies to variables and properties so should only apply there I think.

lib/LanguageServerRename/Model/Renamer.php
dantleech: I mean it will spiral out of control probably. If we focus on a single feature then we can be happy with it faster and merge it, then see about other features.

BladeMF: I am inclined to agree, although I'd be happy to hear the scenario you imagine.

dantleech: Off the top of my head the class moving stuff:

- Rename is moving.
- Moving is moving a file - finding out where that file should live depending on the namespace (Phpactor already has a solution for this problem)
- The current Phpactor implementation moves the file in GiT too (via. the source code filesystem)
- Dealing with class / trait / blah / aliases is always fun.
- The current RPC handler for move will also move any related test files.

There will always be more complexity that will take you by surprise.

For the member (constant,property,method) renaming etc. it may  be simpler, but the logic still differs from local variables considerably (no need to worry about scope etc).

In fact, member renaming may even be the easiest thing, but there again there are subtle differences between the members (as with the `$`)

weeman1337:
This one seems to be replaced by #24 ? Could it then be closed?

dantleech:
yes, this has been extracted and implemented in other PRs :)

PR #22, opened 2021-01-09
Document Locator and Fixing Tests

dantleech:
This PR:

- Adds the TextDocumentLocator abstraction and implementations.
- Prevents the tests from being "killed" by disabling watchers when the test LS is built
- Integrates test implementation of `TestDefinitionLocator`.
- Fixes the stub integration for tests


lib/LanguageServerBridge/LanguageServerBridgeExtension.php
```diff
@@ -25,11 +29,28 @@ public function configure(Resolver $schema)
...
+
+        $container->register(WorkspaceTextDocumentLocator::class, function (Container $container) {
+            return new WorkspaceTextDocumentLocator($container->get(LanguageServerExtension::SERVICE_SESSION_WORKSPACE));
+        });
+
+        $container->register(TextDocumentLocator::class, function (Container $container) {
+            return new ChainDocumentLocator([
+                $container->get(WorkspaceTextDocumentLocator::class),
+                $container->get(FilesystemWorkspaceLocator::class)
+            ]);
```
dantleech, line 53: @BladeMF this is the service which can be used to either get from the workspace all the FS.

tests/LanguageServerCodeTransform/Unit/CodeAction/ImportNameProviderTest.php
```diff
@@ -38,6 +41,7 @@ public function testImportProvider(string $manifest, int $expectedCount, int $ex
 
         $tester->textDocument()->open('file:///foobar', $source);
 
+        wait(delay(10));
```
dantleech: maybe this isn't such a great idea

lib/LanguageServerBridge/LanguageServerBridgeExtension.php
dantleech: or with less typos: to retrieve a document from either the workspace or the filessytem

BladeMF: Which projects changed with this? TextDocument?

dantleech: yes just TextDocument

BladeMF: I need help setting this up for testing. How do I setup an instance of that `WorkspaceTextDocumentLocator` to pass to my renamers?

dantleech: you can use the `InMemoryTextDocumentLocator(TextDocument[])`

dantleech: otherwise something like https://github.com/phpactor/language-server-phpactor-extensions/blob/8e6f7d85dbcff3d812e54508228db6eb11ab8d32/tests/LanguageServerReferenceFinder/Unit/Handler/GotoDefinitionHandlerTest.php#L28-L35

BladeMF: Oh, nice, thanks!

PR #24, opened 2021-01-23
Renamer

BladeMF:
V1 of the rewrite. Currently only [`VariableRenamer`](https://github.com/phpactor/language-server-phpactor-extensions/compare/master...BladeMF:renamer?expand=1#diff-31ec303d27e6f00de525f3995a80910cf546dc683b4045ad3fc775af782d536a) is implemented.

#### Some classes have `2` in the name. 
This is temporary so the previous version of the rename keeps working as I am using it :-)

#### Workflow
Since in LSP a rename is always preceded by a `prepareRename` request which makes all the validation, I've decided to require this flow from the `Renamer2` users. So renamers might collect data on **prepare** which they might use on **rename**.
I expect that when _move class_ is invoked, it will be on some other workspace event (rather than rename), so there should be used the two calls, one after the other `Renamer2::prepareRename` and `Renamer2::rename`.

#### VariableRenamer::locationsToTextEdits
This will probably be moved out of there but currently I can't figure out how to deal with the specifics of class renaming and class moving, i.e.:
* When renaming a class, for each document containing references the `use` statements need to be searched.
* When renaming a class, and a `QualifiedName` is found in a reference, only the last part of the name (containing the class name) must be renamed (including for the `use` statatements)
* When moving a class, this changes part of the namespace of the class, so whenever a `QualifiedName` is found some middle part of it is changed (as opposed to the last part in the case of class renames).

So let's get there and decide then.

#### `NodeUtils2`
This is changed now to return an array of names/tokens since it can be ambiguous which name are you seeking, e.g. the `DefinitionLocator` might return this location:
```php
class MyClass
{
    <>public SomeType $someprop1, SomeType2 $someprop2;
```
There you might want the qualified names `SomeType` and `SomeType2`, and the names `$someprop1` and `$someprop2`. The old version of `NodeUtils` required a the name compare to all the found names, which I decided was a dumb approach.


BladeMF:
The classes that are included in the **rewrite**, are:
[`interface Renamer2`](https://github.com/phpactor/language-server-phpactor-extensions/pull/24/files#diff-1c5b997683f6a80c6ae2eb3bd19c142e760d09c77467c93ba350dbbd3712a11f)
[`class Rename`](https://github.com/phpactor/language-server-phpactor-extensions/pull/24/files#diff-7a62c80f073064ea25903fcd52d3956b81e355a3d41b97f1a6811c055892b161)
[`class RenameLocationGroup`](https://github.com/phpactor/language-server-phpactor-extensions/pull/24/files#diff-ec067996b4a6ae32603c118259d2c306a2880e154b9d72a1c77db82b7523dccf)
[`class RenameLocationsProvider`](https://github.com/phpactor/language-server-phpactor-extensions/pull/24/files#diff-5da58360c1999c08c7176f424a5d4665a0b81bb110690e95debc4646bcd0e1d0)
[`class NodeUtils2`](https://github.com/phpactor/language-server-phpactor-extensions/pull/24/files#diff-ecf9149228b7d6da016ee7ca2f71dcecb61738d564112803e610fe53cbae5388)
[`class VariableRenamer implements Renamer2`](https://github.com/phpactor/language-server-phpactor-extensions/pull/24/files#diff-31ec303d27e6f00de525f3995a80910cf546dc683b4045ad3fc775af782d536a)
[`class RenameResult`](https://github.com/phpactor/language-server-phpactor-extensions/pull/24/files#diff-412f65c2861dcf11b475d3dac6915bf36ec71d37ff82e8ddb1bb241262753c9e)
and the corresponding tests.

lib/LanguageServerRename/Model/Exception/EmptyNewName.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Phpactor\Extension\LanguageServerRename\Model\Exception;
+
+class EmptyNewName extends \Exception
+{
+    public function __construct(string $message = "", int $code = 0, \Throwable $previous = null)
```
dantleech: can remove `$code` and `$previous` and the other instances below.

dantleech (commented):
Looks interesting!

But it is difficult to review as there are multiple things going on and I don't know what is important and what is not. As said before I would really prefer to focus on implementing a single responsibility: variable renaming and through that providing an infrastructure for renaming other things which can be added progressively.

~How does this relate to https://github.com/phpactor/worse-reference-finder/pull/8~ ?

Would it be possible to create a new PR with _only_ the things relating to variable rename and the infrastructure for that? It would be great to have only the minimal code required for this feature.

lib/LanguageServerRename/Model/Renamer/VariableRenamer.php
```diff
@@ -0,0 +1,103 @@
...
+
+    public function prepareRename(TextDocument $textDocument, ByteOffset $offset): ?ByteOffsetRange
+    {
+        $rootNode = $this->parser->parseSourceFile((string)$textDocument);
+        $node = $rootNode->getDescendantNodeAtPosition($offset->toInt());
+        
+        if (
+            (
+                $node instanceof Variable &&
+                null === $node->getFirstAncestor(PropertyDeclaration::class)
```
dantleech, line 58: I don't think this is a variable?

lib/LanguageServerRename/Model/Renamer/VariableRenamer.php
```diff
@@ -0,0 +1,103 @@
...
+            return $range;
+        }
+        return null;
+    }
+    /**
+     * {@inheritDoc}
+     */
+    public function rename(TextDocument $textDocument, ByteOffset $offset, string $newName): Generator
+    {
+        [ $oldName ] = $this->nodeUtils->getNodeNameTexts($this->preparedNode);
```
dantleech: Avoid temporal coupling -- it's fine to have a method `resolveRenameNode` and repeat the work (parser is cached anyway). This method will not work by default given the public API because `preparedNode` is `null`. Other than coupling - this also makes this class stateful.

lib/LanguageServerRename/Model/NodeUtils.php
```diff
@@ -0,0 +1,180 @@
...
+class NodeUtils
+{
+    public function getNodeNameText(Node $node, ?string $name = null): ?string
+    {
+        $token = $this->getNodeNameToken($node, $name);
+        if ($token === null) {
+            return null;
+        }
+        $text = (string)$token->getText($node->getFileContents());
+        if (mb_substr($text, 0, 1) == '$') {
```
dantleech: This is specific to a certain class of rename, and should not be generic.

lib/LanguageServerRename/Model/NodeUtils.php
```diff
@@ -0,0 +1,180 @@
...
+                        && $nodeOrToken->getName() == $name
+                    ) {
+                        return $nodeOrToken->name;
+                    }
+                }
+            }
+            return null;
+        }
+
+        if ($node instanceof ForeachStatement && !empty($name)) {
```
dantleech: when would somebody rename a foreach statement?

dantleech:
With regards to returning an array of tokens from `NodeUtils` - I would probably not have this class handling all the cases. For this PR it would probably be enough to have a Variable token resolver or similar, which can be generalised or not later on.

lib/LanguageServerRename/Model/Exception/EmptyNewName.php
BladeMF: I could probably rebase on master. It's a bit difficult because if I do I lose my rename while I am working on this :-) That is why I did links to all the classes that are involved. Can I help some other way for you to review it while I am working on it?

BladeMF:
Why can't I reply to any of your comments?

lib/LanguageServerRename/Model/NodeUtils.php
BladeMF: That is the old class. Look at `NodeUtils2`.

lib/LanguageServerRename/Model/NodeUtils.php
BladeMF: `foreach($variable1 as $variable => $variable3) {`
These are all variables and are subject to renaming

lib/LanguageServerRename/Model/Exception/EmptyNewName.php
dantleech: Why not check it out independently and work on it as a separate project ?... that's what I do.

lib/LanguageServerRename/Model/Renamer/VariableRenamer.php
BladeMF: I hated to use the same code twice. Thought long about it and decided since the LSP API is coupled I might be as well.

lib/LanguageServerRename/Model/Exception/EmptyNewName.php
BladeMF: I might as well do that... okay.

BladeMF: Why remove `$code` and `$previous`?

lib/LanguageServerRename/Model/NodeUtils.php
BladeMF: Not to mention the case:
```php
$a = new class($var1) { 
    public function __construct($arg1) { .. }
}
```
which I don't think I have covered.

BladeMF:
`NodeUtils2` just returns all the possible names for a location. Returning an array leaves the logic with the caller - 1) get me all the names, 2) let's see which name I am interested in, 3) rename it. Some positions have just one possible name, e.g. `$varia<>ble1` and some have more than one, e.g. `function F(<>Type1 $arg1)`.

lib/LanguageServerRename/Model/NodeUtils.php
dantleech: but `foreach` isn't something you can rename? the variables in the exression are, but if you target them you won't get a foreach, you would, i think, get the `Variable`.

lib/LanguageServerRename/Model/Exception/EmptyNewName.php
dantleech: why wouldn't you? they are unnecessary - why would you pass `$code`, when would you pass `$previous`? it opens up the API unnecessarily. What do `$code` and `$previous` mean in the context of `EmptyNewName`?

dantleech:
> NodeUtils2 just returns all the possible names for a location

yes - _all_ the possible names. but we are only concerned with variables at the moment. i can understand the thinking, but I think that this class provides for 3 distinct operations: `variable rename`, `member rename` and `class-like _moving_` (there is no rename, there is only move). each of which may present their own challenges and require different logic.

BladeMF:
Rebased on master.

BladeMF:
> > NodeUtils2 just returns all the possible names for a location
> 
> yes - _all_ the possible names. but we are only concerned with variables at the moment. i can understand the thinking, but I think that this class provides for 3 distinct operations: `variable rename`, `member rename` and `class-like _moving_` (there is no rename, there is only move). each of which may present their own challenges and require different logic.

I think I get your point and you may be right. Although I still disagree with combining class rename and namespace rename (which is move) as they are nothing like each other.

lib/LanguageServerRename/Model/NodeUtils.php
BladeMF: If I am not mistaken references are returned in such a location that you get the foreach statement rather than the individual variables.

lib/LanguageServerRename/Model/Exception/EmptyNewName.php
BladeMF: Oh, I see. Just this exception you mean.

dantleech:
> Although I still disagree with combining class rename and namespace rename (which is move) as they are nothing like each other.

If you rename a project class (only the short-name), it is, in the common case, a file rename operation or the code no longer works. 

I don't know how the client works with regards to rename, but at least in the current phpactor code you rename a class by changing the FQN or the file path. So I would imagine the *API* accepting an old FQN and a new FQN (as with the [current API](https://github.com/phpactor/class-mover/blob/master/lib/ClassMover.php)). This would also allow us to use the same code with the CLI and legacy RPC handlers.

In fact I wonder if we shouldn't simply update the class-mover library to be compatible with the index :thinking: it could then be used for the LS.

In anycase this is why I'd rather we think about this later.

lib/LanguageServerRename/Model/NodeUtils.php
dantleech: I'm not sure why that would be the case? I would expect the reference to be `foreach (<>$foobar<> as $barfoo) {`

BladeMF:
> > NodeUtils2 just returns all the possible names for a location
> 
> yes - _all_ the possible names. but we are only concerned with variables at the moment. i can understand the thinking, but I think that this class provides for 3 distinct operations: `variable rename`, `member rename` and `class-like _moving_` (there is no rename, there is only move). each of which may present their own challenges and require different logic.

I think I get your point. Although I still disagree with combining class rename and namespace rename (which is move) as they are nothing like each other.

> > Although I still disagree with combining class rename and namespace rename (which is move) as they are nothing like each other.
> 
> If you rename a project class (only the short-name), it is, in the common case, a file rename operation or the code no longer works.
> 
> I don't know how the client works with regards to rename, but at least in the current phpactor code you rename a class by changing the FQN or the file path. So I would imagine the _API_ accepting an old FQN and a new FQN (as with the [current API](https://github.com/phpactor/class-mover/blob/master/lib/ClassMover.php)). This would also allow us to use the same code with the CLI and legacy RPC handlers.
> 
> In fact I wonder if we shouldn't simply update the class-mover library to be compatible with the index 🤔 it could then be used for the LS.
> 
> In anycase this is why I'd rather we think about this later.

I agree they are both FQN changes. Currently you can rename a class by just standing on any of the references or the definition and press F2. It is the same experience as the other renames. I would suppose we can still calculate a new FQN and then call some common code somewhere. Hmm, lets see:

If you rename just the class, you:
- Find all the references + definition
- Change definition name
- Rename file if names match
- For each file with references:
  - Search `use` statements and rename the last part
  - Search references and rename if current name is not aliased. (Probably can skip this if class is found in `use` statements with an alias)
  - If a FQN is found in the references, rename only the last part.

If you rename class' namespace, you:
- Find all the references + definition
- Change the definition namespace
- Move file if old namespace and current calculated namespace are the same
- For each file with references:
  - Search `use` statements and change them as needed.
  - Search references for only FQNs and rename them.

What do you think?

lib/LanguageServerRename/Model/NodeUtils.php
BladeMF: So the locations are:
#### Array
```php
foreach(<>$references as $key=>$value)
```
#### Value
```php
<>foreach($references as $key=>$value)
```

#### Key
```php
<>foreach($references as $key=>$value)
```

BladeMF:
`MemberRenamer` is also done. The code turned out to be identical to `VariableRenamer` except for `getValidNode` and `getNameTokens`.

lib/LanguageServerRename/Model/Renamer.php
```diff
@@ -0,0 +1,15 @@
...
+use Phpactor\TextDocument\ByteOffsetRange;
+use Phpactor\TextDocument\TextDocument;
+
+interface Renamer
+{
+    public function prepareRename(TextDocument $textDocument, ByteOffset $offset): ?ByteOffsetRange;
+    /**
+     * @return \Generator<RenameResult>
+     */
+    public function rename(TextDocument $textDocument, ByteOffset $offset, string $newName): ?\Generator;
```
dantleech, line 14: the `Generator` should not be nullable.

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
```diff
@@ -0,0 +1,182 @@
...
+        TextDocumentLocator $locator,
+        RenameLocationsProvider $locationsProvider
+    ) {
+        $this->parser = $parser;
+        $this->nodeUtils = $nodeUtils;
+        $this->locator = $locator;
+        $this->locationsProvider = $locationsProvider;
+    }
+
+    private function getValidNode(TextDocument $textDocument, ByteOffset $offset): ?Node
```
dantleech, line 54: order methods `public`, `protected` and tinally `private`. this helps promote the `public` API

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function rename(TextDocument $textDocument, ByteOffset $offset, string $newName): Generator
+    {
+        if (($node = $this->getValidNode($textDocument, $offset)) === null) {
+            return null;
+        }
+
+        [ $token ] = $this->getNodeNameTokens($node);
```
dantleech, line 92: this should return a single Token probably.

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
```diff
@@ -0,0 +1,182 @@
...
+        if ($node instanceof MethodDeclaration && $node->name !== null) {
+            return [ $node->name ];
+        }
+
+        if ($node instanceof ScopedPropertyAccessExpression && $node->memberName instanceof Token) {
+            return [ $node->memberName ];
+        }
+        
+        if ($node instanceof Variable && $node->name instanceof Token) {
+            return [ $node->name ];
```
dantleech, line 136: what is this for?

tests/LanguageServerRename/Unit/MemberRenamerTest.php
```diff
@@ -0,0 +1,269 @@
...
+                new class() implements DefinitionLocator {
+                    public function locateDefinition(TextDocument $document, ByteOffset $byteOffset): DefinitionLocation
+                    {
+                        return new DefinitionLocation(TextDocumentUri::fromString(""), ByteOffset::fromInt(0));
+                    }
+                }
+            )
+        );
+        $actualRange = $variableRenamer->prepareRename($document, $selection);
+        $this->assertEquals($expectedRange, $actualRange);
```
dantleech, line 83: replace `$this` with `self`. assertions are static.

dantleech:
What do you think about this:

- `RenameHandle`r: LSP Rename Handler
- `Renamer`: Interface for renaming
- `TolerantRenamer`: Implementation for the `Renamer`:

```
public function rename($document, $offset, $newName): LocationGroups
{
     $node = $this->parser->parse($document)->getDescendantnodeAtPosition($offset);
     $renamer = $this->renamers->renamerForNode($node);
     $locationGroups = $renamer->rename($node, $newName);
     return $locationGroups;
}
```

- `TolerantRenameHandler:` Interface for Tolerant rename handlers (`public function supports($node): bool; public function rename(Node $node, $newName)`)
- `TolerantHandler/VariableRenamer`: The variable renamer.

This would remove the duplication from the renamers.

But if you decide to try this, please open a new PR. In general I think this looks really good, but it's very hard to review when we haven't decided on the essential architecture.

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
BladeMF: We need the array. Property declarations come like this:
```php
<>property string $prop1, $prop2;
```
From this point there are `$prop1` and `$prop2`.

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
BladeMF: When you are standing like this:
```php
property $pro<>p1;
```
The underlying token is a `Variable`. The parents are validated in `getValidToken` - `ScopedPropertyAccessExpression` and `PropertyDeclaration`.

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
dantleech: right, but in this class we only use a single token, the first one?

BladeMF:
I thought we agreed on your proposal from last time. I don't mind they turned out to be similar. I actually quite like getting rid of the if/else in `NodeUtils`. I actually like the current architecture, I just changed it slightly as the `prepareRename` call needs to return a range. That is why there is no `supports` method.

Also the class renamer/mover will probably be nothing like these two. I am beginning to like your idea to do renamers that are usable for both RPC and LSP.

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
BladeMF: Yes? Wait, which method are you referring to that should return a single token?

dantleech: `getNodeNameTokens` => `getNodeNameToken`

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
```diff
@@ -0,0 +1,182 @@
...
+                $this->locationsToTextEdits($locationGroup->uri(), $locationGroup->locations(), $oldName, $newName),
+                $locationGroup->uri()
+            );
+        }
+    }
+    /** @param Location[] $locations */
+    private function locationsToTextEdits(TextDocumentUri $textDocumentUri, array $locations, string $oldName, string $newName): TextEdits
+    {
+        $textEdits = [];
+        $rootNode = $this->parser->parseSourceFile($this->locator->get($textDocumentUri));
```
dantleech, line 107: Note that another strategy would be to not use the locations provided by the reference finder - references from the index can be out-of-date (the watcher could have stopped, it's too busy updating etc).

For the task of renaming members etc it's probably better to just find members with matching names in the given document.,

BladeMF: But members could have been used in other documents?

dantleech: yes absolutely, we would still use the URL - it just seems safer to traverse the node tree rather than relying on the locations. Is there any advantage to using the potentially out-of-date locations?

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
BladeMF: But it can't, as I have demonstrated above :-)

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
BladeMF: Yes, because members need to be resolved, which means invoking worse reflection. There could be a ton of objects in the file using the same name for a method which is not really the same method. I actually think it is out of the question to not use the index. I've been using a version of this code since Christmas and I had bad index only one time. 99% of the time it works perfectly. 

We could probably use the documents from the locations and re-parse all the documents, but that is scary as well.

dantleech: ok then :+1:

BladeMF: as in "keep using the index"?

dantleech: yes

lib/LanguageServerRename/Model/Renamer/MemberRenamer.php
dantleech: Why would we want multiple tokens though? we can only rename one property at a time?

dantleech: and here you only use one....

dantleech: if you are on the point of a property, then what are you renaming? you need to be on the point of a specific property name or?

BladeMF: This is where the reference/definition location is. So yes, I am always using just one, but I don't know which one until I examine all the names at that location:
```php
<>property string $prop1, $prop2;
```
Nearly all of the ambiguity is caused by reference locations at the start of the line/expression:
* `for` statements
* property declarations
* constant declarations

BladeMF: E.g. if we make the locations more precise, we wouldn't need multiple nodes here. It would simplify renaming.

dantleech: right i see., sounds like that's what we should do then finally - I could have a look into that.

BladeMF: ### Property and constant declarations
The only thing I am not sure about is property and constant declarations. In the case below, it is obvious where we want the location:
```php
<>private string $prop1;
```
But in the case of `$prop2` below, I am not sure we don't want it in the same location. 
```php
private string $prop1, $prop2;
```
We would, for example, add an optional `index` to the location. In the case above, it could be `$index=0;` for `$prop1` and `$index=2` for `$prop2`.

### Foreach
Here things are clear, just in front of the variable would be nice.

### List deconstruction
We've missed the case with multi-level arrays (the definition does not work there):
```php
[ [ $var 2 ] ] = function();
```

dantleech: For `private string $prop1, $prop2` we would have the location as `private string $prop1, <>$prop2)`. Ideally also with a range: `<>$prop2</>`

dantleech: Possibly could add some options to the API determining how this works.

dantleech: Anyway, can see the issue now :+1: maybe your option is the one.

dantleech: PHPStorm will do it on the var `<>$prop1` so I think we should update the ref finder to do that at least (if not to include the entire range)

BladeMF: > For private string $prop1, $prop2 we would have the location as private string $prop1, <>$prop2). Ideally also with a range: <>$prop2</>

Yea, that probably is the simplest solution. I am inclined to agree. I wonder if we can check C#/Java behaviour... I'll probably do that and write back.

BladeMF: In C# it selects the name and the cursor is at the end:
```c#
static string _sourcePath<> = @"E:\ERA\Source\";
```

dantleech: yep, it returns the range - i'll see if we can do the same.

weeman1337:
Hey @BladeMF @dantleech as I am desperately waiting for the rename support :D I am ready helping out here. Could you give me a summary of the open points. Following the conversations I didn't completely get it.

dantleech:
@weeman1337 this is mostly now working in `master` - there is support for variables, class names and members. I am currently working on the ability to react on file movements (and so update class references etc)

PR #28, opened 2021-02-07
Renamer: Proper structure

BladeMF:
That is the general structure of the Renamer we talked about. The only thing I changed is the class that managed all the available renamers because I realised that an interface  + a "chain" implementation of the same interface would do it and would be simpler.

The handler is complete with the conversion and stuff and the `ChainRenamer` is implemented. The other stuff is outlined. So let's discuss.


lib/LanguageServerRename/Handler/RenameHandler.php
```diff
@@ -0,0 +1,124 @@
...
+            return new Range(
+                PositionConverter::byteOffsetToPosition(
+                    $range->start(),
+                    (string)$document
+                ),
+                PositionConverter::byteOffsetToPosition(
+                    $range->end(),
+                    (string)$document
+                )
+            );
```
dantleech: Could probably add a utility to convert a Phpactor Range => an LSP Range in the LanguageServerBridge

lib/LanguageServerRename/Model/Renamer.php
```diff
@@ -0,0 +1,15 @@
...
+use Generator;
+use Phpactor\TextDocument\ByteOffset;
+use Phpactor\TextDocument\ByteOffsetRange;
+use Phpactor\TextDocument\TextDocument;
+
+interface Renamer
+{
+    public function getRenameRange(TextDocument $textDocument, ByteOffset $offset): ?ByteOffsetRange;
+    /** @return Generator<RenameResult> */
+    public function rename(TextDocument $textDocument, ByteOffset $offset, string $newName): Generator;
```
dantleech, line 16: not important (and really CS fixer _should_ fix it if it doesn't already) but docblocks should be multiline:

```
/**
 * @return ...
 */
```

and there should be exactly one space between member definitions

tests/LanguageServerRename/Bridge/Unit/ChainRenamerTest.php
```diff
@@ -0,0 +1,179 @@
...
+            {
+                return new ByteOffsetRange(ByteOffset::fromInt(0), ByteOffset::fromInt(1));
+            }
+            
+            public function rename(TextDocument $textDocument, ByteOffset $offset, string $newName): Generator
+            {
+                return;
+                yield;
+            }
+        };
```
dantleech: Rather than using anonymous classes, could we implement a pre defined renamer? (or similar name, TestRenamer, StubRenamer etc)

```
new PredefinedRenamer($byteOffsetRange, $renameResults)
```

This could repalce the stub Member and Variable renamers and potentially make the tests easier?

Possibly you could also implement a builder pattern:

```
PredefinedResult::fromRange(ByteOffsetRange(...))->withResult(...)->withResult(...)
```

This could also be wired into a `TestExtension` for the purposes of integration testing as mentioned below.

tests/LanguageServerRename/Bridge/Unit/ChainRenamerTest.php
```diff
@@ -0,0 +1,179 @@
...
+use Phpactor\TextDocument\ByteOffsetRange;
+use Phpactor\TextDocument\TextDocument;
+use Phpactor\TextDocument\TextDocumentBuilder;
+use Phpactor\TextDocument\TextDocumentUri;
+use Phpactor\TextDocument\TextEdits;
+use function iterator_to_array;
+
+class ChainRenamerTest extends TestCase
+{
+    public function testGetRenameRange_OneMatch(): void
```
dantleech: For the sake of constistency I'd prefer to stick to camel case

tests/LanguageServerRename/Handler/Unit/RenameHandlerTest.php
```diff
@@ -0,0 +1,216 @@
...
+use Phpactor\TextDocument\TextEdits;
+use Prophecy\Argument;
+use Prophecy\PhpUnit\ProphecyTrait;
+use function Amp\Promise\wait;
+
+class RenameHandlerTest extends TestCase
+{
+    use ProphecyTrait;
+
+    public function testPrepareRename(): void
```
dantleech: This would be better as an integration test .... you could add an additional extension with the `PredefinedRenamer` configured as menitoned above. See the tests for the other handlers.

dantleech (commented):
Looks good! I think it might be better to implement a `PredefinedRenamer` (or however it should be named) as mentioned above. This will help with testing in general (including manual tests)

tests/LanguageServerRename/Handler/Unit/RenameHandlerTest.php
dantleech: See here for a similar approach https://github.com/phpactor/language-server-phpactor-extensions/blob/35c2564e7fc092bc417bdee230a3c0583caac0b3/tests/LanguageServerIndexer/Extension/TestExtension.php#L14

BladeMF:
> Looks good! I think it might be better to implement a `PredefinedRenamer` (or however it should be named) as mentioned above. This will help with testing in general (including manual tests)

What do you mean `PredefinedRenamer`?

lib/LanguageServerRename/Handler/RenameHandler.php
BladeMF: Agreed. Was thinking about that.

tests/LanguageServerRename/Bridge/Unit/ChainRenamerTest.php
BladeMF: Sounds good.

tests/LanguageServerRename/Bridge/Unit/ChainRenamerTest.php
BladeMF: That's a bit of "uncle bob" thing. Sure.

lib/LanguageServerRename/Handler/RenameHandler.php
BladeMF: Do you want _me_ to do that or will you do it?

lib/LanguageServerRename/Model/Renamer.php
BladeMF: You know, I've been missing help with this. I've used extensions which when inside `/**` and you hit enter, you get in the proper indent with ` * ` already typed, getting to:
```php
    /**
     * <>
```

lib/LanguageServerRename/Handler/RenameHandler.php
dantleech: Up to you, feel free to leave it.

BladeMF: Yea, I probably should do it then....

BladeMF:
Interestingly enough I can't reproduce the fail of php 7.4 unit tests. I must be missing something.

lib/LanguageServerBridge/Converter/RangeConverter.php
```diff
@@ -0,0 +1,17 @@
...
+class RangeConverter
+{
+    public static function toLspRange(ByteOffsetRange $range, string $text): Range
+    {
+        return new Range(
+            PositionConverter::byteOffsetToPosition($range->start(), $text),
+            PositionConverter::byteOffsetToPosition($range->end(), $text),
+        );
+    }
+}
```
dantleech, line 17: :+1:

dantleech:
:boom:

PR #29, opened 2021-03-06
Variable renamer

BladeMF:
That's only the variable renamer. 

Currently the integration test is failing because I am unsure what to do - it seems like only `InMemoryRenamer` is supposed to be in the container, but now there's the `VariableRenamer`, `RenameLocationsProvider` in there as well. So @dantleech some advice on this would be appreciated.

### DIFF
--- a/bin/serve
+++ b/bin/serve
@@ -55,10 +55,8 @@ $container = PhpactorContainer::fromExtensions([
     LanguageServerBridgeExtension::class,
     LanguageServerSelectionRangeExtension::class,
     LanguageServerCodeTransformExtension::class,
-
     LanguageServerRenameExtension::class,
     LanguageServerWorseRenameExtension::class,
-
     WorseReferenceFinderExtension::class,
     CodeTransformExtension::class,
     PhpExtension::class,
--- a/bin/serve
+++ b/bin/serve
@@ -17,6 +17,8 @@ use Phpactor\Extension\LanguageServerCompletion\LanguageServerCompletionExtensio
 use Phpactor\Extension\LanguageServerHover\LanguageServerHoverExtension;
 use Phpactor\Extension\LanguageServerIndexer\LanguageServerIndexerExtension;
 use Phpactor\Extension\LanguageServerReferenceFinder\LanguageServerReferenceFinderExtension;
+use Phpactor\Extension\LanguageServerRename\LanguageServerRenameExtension;
+use Phpactor\Extension\LanguageServerRename\LanguageServerWorseRenameExtension;
 use Phpactor\Extension\LanguageServerSelectionRange\LanguageServerSelectionRangeExtension;
 use Phpactor\Extension\LanguageServerSymbolProvider\LanguageServerSymbolProviderExtension;
 use Phpactor\Extension\LanguageServerWorseReflection\LanguageServerWorseReflectionExtension;
@@ -26,6 +28,7 @@ use Phpactor\Extension\Logger\LoggingExtension;
 use Phpactor\Extension\Php\PhpExtension;
 use Phpactor\Extension\ReferenceFinder\ReferenceFinderExtension;
 use Phpactor\Extension\SourceCodeFilesystem\SourceCodeFilesystemExtension;
+use Phpactor\Extension\WorseReferenceFinder\WorseReferenceFinderExtension;
 use Phpactor\Extension\WorseReflection\WorseReflectionExtension;
 use Phpactor\FilePathResolverExtension\FilePathResolverExtension;
 use Phpactor\Indexer\Extension\IndexerExtension;
@@ -52,6 +55,11 @@ $container = PhpactorContainer::fromExtensions([
     LanguageServerBridgeExtension::class,
     LanguageServerSelectionRangeExtension::class,
     LanguageServerCodeTransformExtension::class,
+
+    LanguageServerRenameExtension::class,
+    LanguageServerWorseRenameExtension::class,
+
+    WorseReferenceFinderExtension::class,
     CodeTransformExtension::class,
     PhpExtension::class,
 ], [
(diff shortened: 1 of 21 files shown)


bin/serve
```diff
@@ -52,6 +55,11 @@ $container = PhpactorContainer::fromExtensions([
     LanguageServerBridgeExtension::class,
     LanguageServerSelectionRangeExtension::class,
     LanguageServerCodeTransformExtension::class,
+
+    LanguageServerRenameExtension::class,
+    LanguageServerWorseRenameExtension::class,
+
```
dantleech: probably shouldn't have white space here...

lib/LanguageServerRename/Handler/RenameHandler.php
```diff
@@ -105,20 +105,21 @@ private function resultToWorkspaceEdit(iterable $results): WorkspaceEdit
         $documentChanges = [];
```
dantleech: this var is not used

tests/LanguageServerRename/OffsetExtractor.php
```diff
@@ -0,0 +1,94 @@
...
+    private $pointCreateors = [];
+
+    /** @var array */
+    private $rangeOpenMarkers = [];
+    /** @var array */
+    private $rangeCloseMarkers = [];
+    /** @var Closure[] */
+    private $rangeCreateors = [];
+
+    public function registerPoint(string $name, string $marker, Closure $creator = null): void
```
dantleech: can we change this to not have a factory, but instead always return `ByteOffset` and `ByteOffsetRange` as relevant.

tests/LanguageServerRename/OffsetExtractor.php
```diff
@@ -0,0 +1,94 @@
...
+            if (isset($this->rangeOpenMarkers[$result])) {
+                $currentRangeStartOffset = $offset;
+                continue;
+            }
+
+            if (isset($this->rangeCloseMarkers[$result])) {
+                if (!isset($retVal[$this->rangeCloseMarkers[$result]])) {
+                    $retVal[$this->rangeCloseMarkers[$result]] = [];
+                }
+                $retVal[$this->rangeCloseMarkers[$result]][] = $this->rangeCreateors[$result]($currentRangeStartOffset, $offset, $newSource);
```
dantleech: typo!

tests/LanguageServerRename/OffsetExtractor.php
```diff
@@ -0,0 +1,94 @@
+<?php
+
+namespace Phpactor\Extension\LanguageServerRename\Tests;
+
+use Closure;
+
+class OffsetExtractor
```
dantleech, line 9: could create this with a static constructor e.g. `::create()` and then a fluent API `::create()->registerX()->registerY()`

BladeMF:
How shall we proceed with the failing test? Currently on PHP8 only ImportNameProviderTest is failing.

dantleech:
I think it's flaky - restarted

dantleech:
For the extractor I was thinking:

```
$extractor = ... // create extractor
$range = $extractor->range('foobar'); // public function range($name): ByteOffsetRange
$selection = $extractor->offsets('.') // /** @return ByteOffset[] */public function selection($name): array
```

or similar - as it's more explicit and type safe. WDYT?

BladeMF:
Is this the registration or the results?
________________________________

BladeMF:
If we have:
```
$extractor = OffsetExtractor::create()
  ->registerPoint(...)
  ->registerRange(....)
  ->parse($source);
```
then probably will be better:
```
$point = $extractor->getPoint(...);
$range = $extractor->getRange(...);
```
what do you think?

dantleech:
yeah I think that makes sense - wasn't sure if it would be `point()` or `points()`

(i would drop the `get` prefix - mostly because that's the idiom in phpactor)

we can have both point and points etc as required

where `point(): ByteOffset` would throw an exception if the point wasn't found

tests/LanguageServerRename/OffsetExtractor.php
```diff
@@ -82,20 +94,41 @@ public function parse(string $source): array
...
+
+        return $this;
+    }
+    /**
+     * @return ByteOffset[]
+     */
+    public function points(string $name): array
+    {
+        if(!isset($this->pointResults[$name]))
+            throw new RuntimeException('No registered point by that name.');
```
dantleech: This would be more helpful when developing: `sprintf('No point registered with name "%s", known names "%s"', $name, implode('", "', array_keys($this->pointResults)))`

tests/LanguageServerRename/OffsetExtractor.php
```diff
@@ -82,20 +94,41 @@ public function parse(string $source): array
...
-        $retVal['newSource'] = $newSource;
+        return $this->pointResults[$name];
+    }
+    /**
+     * @return ByteOffsetRange[]
+     */
+    public function ranges(string $name): array
+    {
+        if(!isset($this->rangeResults[$name]))
+            throw new RuntimeException('No registered range by that name.');
```
dantleech: same

BladeMF:
Agreed, not sure what I was thinking with such generic text.

Vladi.
________________________________

tests/LanguageServerRename/OffsetExtractor.php
```diff
@@ -42,7 +56,7 @@ public function registerRange(string $name, string $openMarker, string $closeMar
         return $this;
     }
 
-    public function parse(string $source): array
+    public function parse(string $source): OffsetExtractor
```
dantleech, line 59: I would consider making this the static constructor `create($source)`, replacing the existing constructor. This would make the class immutable. The disadvantage is that we can't reuse the config.

Another option would be to return a new object which just has the points and ranges - so this method then becomes a factory..

But, let's go with this unless you want to try changing it. In anycase, I think it can be refactored without changing the API which is the main thing

BladeMF:
I'm 50/50. create($source)​ is simpler to use. Let me think for a bit.

Vladi.
________________________________

dantleech:
Cheers @BladeMF , I'll update the extractor

PR #32, opened 2021-04-11
Offset Extractor imporvements

dantleech:
- Extracts an `OffsetExtractorResult` (the offset extractor becomes a factory for this).
- Allow accessing point/ranges without a name `OffsetExtract::extractOffset($source)->point()`

### DIFF
--- a/tests/LanguageServerRename/Util/OffsetExtractor.php
+++ b/tests/LanguageServerRename/Util/OffsetExtractor.php
@@ -20,18 +20,9 @@ final class OffsetExtractor
      */
     private $rangeCloseMarkers = [];
 
-    final private function __construct()
-    {
-    }
-
     public static function create(): OffsetExtractor
     {
         return new OffsetExtractor();
-    }
-
-    public static function extractOffset(string $source, string $marker = '<>'): OffsetExtractorResult
-    {
-        return self::create()->registerOffset('point', $marker)->parse($source);
     }
 
     public function registerOffset(string $name, string $marker): OffsetExtractor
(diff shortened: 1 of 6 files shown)


tests/LanguageServerRename/Util/OffsetExtractor.php
```diff
@@ -20,21 +19,8 @@ class OffsetExtractor
...
-     */
-    private $rangeResults = [];
-    /**
-     * @var string
-     */
-    private $source;
-
 
-    private function __construct()
+    final private function __construct()
```
dantleech: redundant as class is already final, remove the constructor.

tests/LanguageServerRename/Util/OffsetExtractor.php
```diff
@@ -43,7 +29,12 @@ public static function create(): OffsetExtractor
         return new OffsetExtractor();
     }
 
-    public function registerPoint(string $name, string $marker): OffsetExtractor
+    public static function extractOffset(string $source, string $marker = '<>'): OffsetExtractorResult
+    {
+        return self::create()->registerOffset('point', $marker)->parse($source);
+    }
```
dantleech: this convenience method isn't tested

tests/LanguageServerRename/Util/OffsetExtractor.php
```diff
@@ -56,7 +47,7 @@ public function registerRange(string $name, string $openMarker, string $closeMar
         return $this;
     }
 
-    public function parse(string $source): OffsetExtractor
+    public function parse(string $source): OffsetExtractorResult
```
dantleech, line 41: note that this method no longer modifies the extractor's state, it produces a result.

dantleech:
cc @BladeMF

PR #33, opened 2021-04-11
Member rename

dantleech:
Implements member renaming

- Changes the Rename API to return `Generator<LocatedTextEdit>` (which is a URI and a TextEdit): the caller is responsible for aggregating them now.
- Introduces (via. reference-finder package) the `DefinitionAndReferenceFinder` which can be used to remove the need to inhect both of these,


lib/LanguageServerRename/Adapter/Worse/MemberRenamer.php
```diff
@@ -0,0 +1,127 @@
...
+            yield new LocatedTextEdit(
+                $reference->location()->uri(),
+                PhpactorTextEdit::create(
+                    $range->start(),
+                    $range->end()->toInt() - $range->start()->toInt(),
+                    $newName
+                )
+            );
+        }
+    }
```
dantleech: note that this method is not specific to members. We could probably change the variable renamer to use exactly this logic (e.g. one solution would be having an abstract "reference finder renamer" class). Class renaming would also be very similar.

lib/LanguageServerRename/Adapter/Worse/RenameLocationsProvider.php
```diff
@@ -9,6 +9,9 @@
 use Phpactor\TextDocument\ByteOffset;
 use Phpactor\TextDocument\TextDocument;
 
+/**
+ * @deprecated The renamers should stream LocatedTextEdit instances.
+ */
 class RenameLocationsProvider
```
dantleech: these classes a redundant with the new refactoring, but I'm too tired to factor them out at the moment.

lib/LanguageServerRename/LanguageServerRenameWorseExtension.php
```diff
@@ -43,6 +39,19 @@ public function load(ContainerBuilder $container): void
         }, [
             LanguageServerRenameExtension::TAG_RENAMER => []
         ]);
+
+        $container->register(MemberRenamer::class, function (Container $container) {
+            return new MemberRenamer(
+                new DefinitionAndReferenceFinder(
```
dantleech, line 41: this class combines the definition finder with the reference finder, so that it includes the definition in the references

lib/LanguageServerRename/Adapter/Worse/MemberRenamer.php
BladeMF: They are all very, very similar, but not quite. We can probably work out a way to group them though.

lib/LanguageServerRename/Adapter/Worse/RenameLocationsProvider.php
BladeMF: Did you change the reference finder to return the definition?

lib/LanguageServerRename/Adapter/Worse/MemberRenamer.php
```diff
@@ -0,0 +1,127 @@
...
+        Parser $parser
+    ) {
+        $this->referenceFinder = $referenceFinder;
+        $this->locator = $locator;
+        $this->parser = $parser;
+    }
+
+    public function getRenameRange(TextDocument $textDocument, ByteOffset $offset): ?ByteOffsetRange
+    {
+        $node = $this->parser->parseSourceFile($textDocument->__toString())->getDescendantNodeAtPosition($offset->toInt());
```
BladeMF: The method is called for each reference, we can be parsing the same file multiple times. I think we still need the "group by document" logic.

lib/LanguageServerRename/Adapter/Worse/RenameLocationsProvider.php
dantleech: No, I introduced a ReferenceFinder which uses both a definition locator and a reference finder, so kindof.

lib/LanguageServerRename/Adapter/Worse/MemberRenamer.php
dantleech: The calling class is responsible for grouping now, so `LocatedTextEditMap` collects all the location/edit objects and groups them by document.

dantleech: hm, sorry I see your point now. The parser is cached, so it shouldn't affect performance, but there's a better way,

lib/LanguageServerRename/Adapter/Worse/MemberRenamer.php
dantleech: I think the variable renamer could use the same logic if the variable definition locator behaved as expected (e.g. `<>foreach ($foo` is located instead of `foreach (<>$foo`.

Renaming classes is a bit more involved as it requires to change the use statements.

lib/LanguageServerRename/Handler/RenameHandler.php
```diff
@@ -99,11 +101,12 @@ public function registerCapabiltiies(ServerCapabilities $capabilities): void
         $capabilities->renameProvider = new RenameOptions(true);
     }
     
-    private function resultToWorkspaceEdit(iterable $results): WorkspaceEdit
+    private function resultToWorkspaceEdit(Traversable $results): WorkspaceEdit
     {
         $documentEdits = [];
-        foreach ($results as $result) {
-            /** @var RenameResult $result */
+        $map = LocatedTextEditsMap::fromLocatedEdits(iterator_to_array($results));
```
dantleech: this will be blocking, use `foreach ($results as $result)` and potentially adda `delay(1)` every so often to unblock.

lib/LanguageServerRename/Adapter/Worse/MemberRenamer.php
dantleech: Saying that, the cache should be sufficient,

lib/LanguageServerRename/Adapter/Worse/VariableRenamer.php
```diff
@@ -2,194 +2,27 @@
...
 use Phpactor\TextDocument\ByteOffsetRange;
-use Phpactor\TextDocument\Location;
-use Phpactor\TextDocument\TextDocument;
-use Phpactor\TextDocument\TextDocumentLocator;
-use Phpactor\TextDocument\TextDocumentUri;
-use Phpactor\TextDocument\TextEdit;
-use Phpactor\TextDocument\TextEdits;
 
-class VariableRenamer implements Renamer
+class VariableRenamer extends AbstractReferenceRenamer
```
dantleech, line 11: @BladeMF I fixed the variable definition locator, now both the member and variable renamers differ only with the detection of the range

lib/LanguageServerRename/LanguageServerRenameWorseExtension.php
dantleech: And now it's not necessary as the variable reference finder will also find the definition (when passed a flag in the constructor)

tests/LanguageServerRename/Unit/Adapter/Worse/VariableRenamerTest.php
```diff
@@ -152,39 +146,52 @@ public function provideRename(): Generator
...
         ];
 
         yield 'Rename variable (as foreach value)' => [
-            '<?php class Class1 { function method1(){ $var = []; <d>foreach($var as ${{val}}) { <r>${{v<>al}} += 5; } } }'
+            '<?php class Class1 { function method1(){ $var = []; foreach($var as <d>${{val}}) { <r>${{v<>al}} += 5; } } }'
         ];
 
         yield 'Rename variable (as foreach key)' => [
-            '<?php class Class1 { function method1(){ $var = []; <d>foreach($var as ${{key}}=>$val) { <r>${{k<>ey}} += 5; } } }'
+            '<?php class Class1 { function method1(){ $var = []; foreach($var as <d>${{key}}=>$val) { <r>${{k<>ey}} += 5; } } }'
```
dantleech, line 169: the fixed definition finder locates the var correctly

lib/LanguageServerRename/Adapter/Worse/VariableRenamer.php
BladeMF: Good, that was _one_ solution. I shied away as I am not certain that all renamers will have this logic, but calling it `ReferenceRenamer` is probably a good strategy!

lib/LanguageServerRename/Adapter/Worse/MemberRenamer.php
BladeMF: It has cache? I didn't know that. Probably there is no reason to group them here now.

lib/LanguageServerRename/Adapter/Worse/MemberRenamer.php
BladeMF: I am really for the proper location of references. It gave me great headache.

lib/LanguageServerRename/Handler/RenameHandler.php
BladeMF: Currently I haven't had a slow rename, so not sure it will be necessary.

dantleech: Try looking for a method which like `next()` in a big project, it will probably time out after 10 seconds. Anything that is amiguous can take a long time.

PR #35, opened 2021-04-23
Replace completion command for imports with additionalTextEdits

### LINKED ISSUE #1248: LSP Class / function imports should be provided by additionalTextEdits instead as a command
At the moment class and function imports are provided as a command with the completion item: https://github.com/phpactor/language-server-phpactor-extensions/blob/master/lib/LanguageServerCompletion/Handler/CompletionHandler.php#L123

My completion kind of freaks out because at the same time the completion is inserted the command modifies the document. Digging the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion) I found that it is suggested to provide imports via `additionalTextEdits`:

> Additional text edits should be used to change text unrelated to the
> current cursor position (for example adding an import statement at the
> top of the file if the completion item will insert an unqualified type).

Looking into the completion plugins I found that they support `additionalTextEdits` as well.

As I already found all the places I could work on this one if the approach is okay @dantleech .

weeman1337:
This PR replaces the completion command for imports with `additionalTextEdits`.

Summary
* new `ImportName` service handling all the things the import command did before
* the import command and the completion now utilise the `ImportName` service for creating the import text edits

Closes phpactor/phpactor#1248


tests/LanguageServerCompletion/Extension/TestExtension.php
```diff
@@ -0,0 +1,30 @@
...
+        $this->registerFilePathExpanders($container);
+    }
+
+    public function configure(Resolver $schema): void
+    {
+    }
+
+    private function registerFilePathExpanders(ContainerBuilder $container): void
+    {
+        $container->register('core.file_path_resolver.project_config_expander', function (Container $container) {
```
weeman1337, line 25: This one comes from the phpactor main project. I've put this here for a minimal setup.

weeman1337:
@dantleech there is an issue when completing the same simple names:

ImportConflicts

But I think this is a separate issue (is this a bug or a new feature)? I would expect that it uses the FQN if there is already an import for the same simple name.

dantleech:
eh ... this is something the current import logic is taking care of - if there is a duplicate name it will automatically add an alias https://github.com/phpactor/language-server-phpactor-extensions/blob/320b8b0e1103a648f1f6a1892f5604e7c964252d/lib/LanguageServerCodeTransform/LspCommand/ImportNameCommand.php#L75-L99

weeman1337:
I think then it should behave the same for completions as well.

Should that be part of the `CompletionHandler`? Like

* Using parts of `TolerantImportName` (adding an option not to update references)
  * Instead of using the `TolerantUpdater` as in the current PR
* Extract the alias fallback logic from `ImportNameCommand.php` and reuse it there

dantleech:
Could we use the exact same logic? It seems maybe we could also delegate to the command in-process?

weeman1337:
> Could we use the exact same logic?

That was what I was trying to say above :D 

> It seems maybe we could also delegate to the command in-process?

I will try. But the imported name has to be changed as well. If the command introduces the imports then there has to be feedback to the completion to change the completion text.

dantleech:
Yes - maybe a new API method on `ImportName` or, as you say, a flag.

We could then:

1. ~Delegate to the command in-line (I can check what the best way to do that would be) and add a flag to the command.~
2. Extract the logic from the command (and the tests surrounding that logic) to a new service in the `[LS]CodeTransform` package.
3. Refactor the command into a service - I'm not sure it's used for anything else, so removing some responsibility there might be good.

~In the case of the command we would still need to have access to the new completion suggestion if an Alias was used. So perhaps the best option is 2)~ (as you say :sweat_smile: )

lib/LanguageServerNameImport/Service/NameImport.php
```diff
@@ -0,0 +1,90 @@
+<?php
```
weeman1337: @dantleech I extracted the import functionality to this "service". Could you please have a look whether you are fine with the naming and location of it? Thanks :)

(Using this new service is still open. See ToDo in the PR description.)

lib/LanguageServerNameImport/LanguageServerNameImportExtension.php
```diff
@@ -0,0 +1,31 @@
...
+
+use Phpactor\CodeTransform\Domain\Refactor\ImportName;
+use Phpactor\Container\Container;
+use Phpactor\Container\ContainerBuilder;
+use Phpactor\Container\Extension;
+use Phpactor\Extension\LanguageServer\LanguageServerExtension;
+use Phpactor\Extension\LanguageServerNameImport\Service\NameImport;
+use Phpactor\MapResolver\Resolver;
+
+class LanguageServerNameImportExtension implements Extension
```
dantleech: Instead of creating another extension, can this go in the `code-transform` extension?

lib/LanguageServerNameImport/Service/NameImport.php
dantleech: Could put it in `Model` I guess (it's both code-transform and language-server so nothing "extrenal" really)

dantleech: If we move this to `LanguageServerCodeTransform` additionally some namespacing `Model\ImprotName`

tests/LanguageServerCompletion/Extension/TestExtension.php
dantleech: hmm, yes this is annoying. Perhaps the container could somehow take care of path expansion (like Symfony) the reason I did not do this originally was performance for the original Phpactor which is a new process for each RPC request - could use a cache :/

tests/LanguageServerNameImport/Service/NameImportServiceTest.php
```diff
@@ -0,0 +1,271 @@
...
+        $this->importNameProphecy = $this->prophesize(RefactorImportName::class);
+        $this->byteOffset = ByteOffset::fromInt(self::EXAMPLE_OFFSET);
+        $this->aliasAlreadyUsedExceptionProphecy = $this->prophesize(AliasAlreadyUsedException::class);
+        $this->nameAlreadyImportedExceptionProphecy = $this->prophesize(
+            NameAlreadyImportedException::class
+        );
+
+        $this->textEdits = TextEdits::one(
+            TextEdit::create(23, 6, 'huhuhu')
+        );
```
dantleech: you can avoid prophecy in most of these cases:

- for the value object `TextDocumentItem` use `ProtocolFactory::textDocumentItem()` (which is a convenient wrapper for common protocol classes for testing)
- the Exception doesn't need to be prophesized
- You can use the real `Workspace` class I think.

dantleech (commented):
Really nice work so far :+1:

tests/LanguageServerNameImport/Service/NameImportServiceTest.php
weeman1337: My idea was to isolate the subject under test as good as possible.  
That is why I prophecised / mocked everything possible except `NameImport`.

Maybe just different opinions about test strategies. I am fine with either ways.

lib/LanguageServerNameImport/LanguageServerNameImportExtension.php
weeman1337: Yes. `code-transform` needs some changes anyway.

tests/LanguageServerNameImport/Service/NameImportServiceTest.php
dantleech: yes - it's a concious choice to move away from doing this. Mocked dependencies do not act in the same way as the real objects, and some objects absolutely do not need to be mocked. Ideally we'd always test using real things to get the most value from a test, but practically we need to use fakes, stubs and mocks sometimes.

It might make sense to mock something with an _interface_ and complicated behavior and external dependencies (e.g. the `NameImport`) but things that belong to the domain we're testing needn't be mocked. And some types of class should never be mocked (Value Objects, Exceptions, etc). We're testing behavior.

lib/LanguageServerNameImport/LanguageServerNameImportExtension.php
weeman1337: With `code-transform" extension do you mean https://github.com/phpactor/language-server-phpactor-extensions/tree/master/lib/LanguageServerCodeTransform or https://github.com/phpactor/code-transform ?

dantleech: yeah, i mean the LanguageServerCodeTransform extension in this repo.

theoretically this package offers extensions which bridge the language server to phpactor's functionality. so `name import` is a `code-transform` thing, so should belong there.

composer.json
```diff
@@ -13,7 +13,7 @@
         "php": "^7.3 || ^8.0",
         "dantleech/object-renderer": "^0.1.1",
         "phpactor/phly-event-dispatcher": "^2.0.0",
-        "phpactor/code-transform": "^0.4.1",
+        "phpactor/code-transform": "dev-feature-cmd-addtextedits as 0.4.2",
```
weeman1337, line 16: - [ ]  Rollback once the `code-transform` changes have been merged

lib/LanguageServerCompletion/Handler/CompletionHandler.php
```diff
@@ -111,16 +111,25 @@ public function completion(CompletionParams $params, CancellationToken $token):
                         : InsertTextFormat::PLAIN_TEXT
                     ;
                 }
-                
-                $items[] = $i = CompletionItem::fromArray([
+
+                $importNameResult = $this->importClassOrFunctionName($suggestion, $params);
+
+                if ($importNameResult->isSuccessAndHasAliasedNameImport()) {
+                    // todo is this the right place?
```
weeman1337: My feelings say this is the wrong place and way for doing this.  
See the "text edit conversion" comment below as well.

lib/LanguageServerCompletion/Handler/CompletionHandler.php
```diff
@@ -111,16 +111,25 @@ public function completion(CompletionParams $params, CancellationToken $token):
...
+                }
+
+                $items[] = CompletionItem::fromArray([
                     'label' => $name,
                     'kind' => PhpactorToLspCompletionType::fromPhpactorType($suggestion->type()),
                     'detail' => $this->formatShortDescription($suggestion),
                     'documentation' => $suggestion->documentation(),
                     'insertText' => $insertText,
                     'sortText' => $this->sortText($suggestion),
+                    // todo convert text edit if there is an alias import
```
weeman1337: The imported name and the text edit have to be changed for an alias import. I am unsure whether this belongs to the `CompletionHandler`. But I have no idea where to put it otherwise.

lib/LanguageServerCompletion/Handler/CompletionHandler.php
```diff
@@ -166,20 +175,31 @@ private function textEdit(Suggestion $suggestion, TextDocumentItem $textDocument
...
+            return ImportNameResult::createEmptyResult();
         }
 
-        return new Command(
-            'Import class',
-            ImportNameCommand::NAME,
-            [$uri, $offset->toInt(), $suggestion->type(), $suggestion->nameImport()]
+        $doc = $this->workspace->get($params->textDocument->uri);
+        // todo: avoid conversions?
+        $offset = PositionConverter::positionToByteOffset($params->position, $doc->text);
```
weeman1337, line 192: I found several places where positions are converted. After all I found three options: Integer, ByteOffset and Position. Maybe not a task for this PR but in general it maybe would make sense to replace all of these by an unified Position providing line/col and the byte offset. That could save many conversion calls.

weeman1337:
* The name import logic is now located in `LanguageServerCodeTransform`.
* The tests now do less mocking.
* The completion now uses the name import logic as well.
* Ignore the failed tests for now. This is because of my temporary composer dependency change. I will rollback the composer changes once the `code-transform` PR phpactor/code-transform#50 has been merged.

I still have some open points where I do not really know what to do. See the code comments.

lib/LanguageServerCompletion/Handler/CompletionHandler.php
dantleech: In general Phpactor is not coupled to LSP (for historical reasons) so this is where we cross domains. Phpactor has it's own value objects in `phpactor/text-document`: `LineCol` for example translates to LSP `Position`. We could imagine one day migrating older Phpactor code to use the new Phpactor V/Os which correspond more or less with the LSP ones.

On the positive side the Phpactor V/Os are much more useful to work with, as they can provide utility methods, whereas the LSP protocol classes are basically DTOs.

lib/LanguageServerCompletion/Handler/CompletionHandler.php
dantleech: Could add a bridge/adapter to avoid the coupling, I think the smell is that we import classes from CodeTransform here, you could introduce a `Model\CompletionNameImporter` interface (or whatever) and then add an adapter in `LanugageServerCompletion/Adapter/LanguageServerCodeTransform/CodeTransformCompletionNameImporter` for example.

What do you think? Would that help?

weeman1337: That doesn't solve the problem:
* The completion provides suggestions (e.g. `Foo`)
* The import function finds out that it has to be imported using an alias name (e.g. `BarFoo`)
* Now this has to be feed back to the completion because the imported name is now different from the original suggestion

dantleech: Why does it need to be fed back? Is there a functional problem with the code as it is?

weeman1337: Because the import may result in using an alias. As described above: The original suggestion is `Foo`. Because of the import an alias has to be used. `Foo` becomes for e.g. `BarFoo`. If that happens the completion `insertText` and the `textEdit` have to be changed from `Foo` to `BarFoo`. This is easy for the `insertText` but not for the `textEdit`.

weeman1337: To be concrete: this has to change if there is an alias: https://github.com/weeman1337/language-server-phpactor-extensions/blob/78b68789ff221cd084084d6e5156a0c3e071e3df/lib/LanguageServerCompletion/Handler/CompletionHandler.php#L157-L176

lib/LanguageServerCompletion/Model/CompletionNameImporter.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phpactor\Extension\LanguageServerCompletion\Model;
+
+use Phpactor\Extension\LanguageServerCodeTransform\Model\NameImporter\NameImporterResult;
+
+interface CompletionNameImporter
```
weeman1337, line 9: I hope that is what you expected for the decoupling @dantleech  https://github.com/phpactor/language-server-phpactor-extensions/pull/35#discussion_r632349702

lib/LanguageServerCompletion/Handler/CompletionHandler.php
weeman1337: Hai @dantleech . Just a ping if you have any ideas about how to solve the problem mentioned above.

dantleech: I'm sure I'm missing something: but why not refactor to pass the `$importName` in addition to `Suggestion`? The range remains constant? Only the replacement text needs to change?

weeman1337: I am not sure if I am proably missing something 🙃

I had a closer look at the `TextEdit` spec. First I thought that it contains the target range. That isn't the case. I will update the code. It should work.

weeman1337:
Great - with https://github.com/phpactor/language-server-phpactor-extensions/commit/db5c6c4474ab93059bd33d82b9cf871ab357f298 some parts of this PR are done as well but in a slightly different way. I will close this one and open some smaller PRs.

PR #36, opened 2021-04-24
Allow hover template override

BladeMF:
Add a config setting `language_server_hover.template_paths` that allows the user to add more paths (other than `__DIR__ . '/../../templates/markdown'`) that are searched for twig templates for the hover content.

### DIFF
--- a/lib/LanguageServerHover/LanguageServerHoverExtension.php
+++ b/lib/LanguageServerHover/LanguageServerHoverExtension.php
@@ -17,8 +17,9 @@ use Phpactor\MapResolver\Resolver;
 
 class LanguageServerHoverExtension implements Extension
 {
+    public const PARAM_TEMPLATE_PATHS = 'language_server_hover.template_paths';
+    
     private const SERVICE_MARKDOWN_RENDERER = 'language_server_completion.object_renderer.markdown';
-    private const PARAM_TEMPLATE_PATHS = 'language_server_hover.template_paths';
 
     /**
      * {@inheritDoc}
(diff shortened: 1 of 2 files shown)


dantleech:
Instead of allowing this, can we also / instead fix the broken offset hover template? or what is the issue with the current templates before merding this?

BladeMF:
> Instead of allowing this, can we also / instead fix the broken offset hover template? or what is the issue with the current templates before merding this?

The formatting wasn't OK - missing newlines, the source code view wasn't colourised after the first line, etc. I suppose VS Code interprets markdown a bit different. So wasn't just the frames.

Also, why not allow customisation? This seems like one of those personal things where everyone likes it different. I played with them quite a bit and ended up simplifying them.

dantleech:
Not against it, just wondering if it was only f.e. to fix the ugly "offset" hover, or if there are other issues

dantleech:
Anyway, build it failing with some new hell. You probably need to copy the `project_config` configuration from one of the other container builders in another extensions.

BladeMF:
I'll try and give examples:

1) No colouring beyond the first line:
image
2) Missing new lines in frames:
image

3) Not terribly fond of that format:
image
* it is missing the inferred return type
* The full namespace of the owner is kind of confusing

4) Also, the missing links for the built-in PHP functions was the first thing that made me go and look. Since I couldn't find a parsed Doc block object, I just put it as a string. It's not pretty, but I have my links (they are present as @link element in the phpstorm stubs)

BladeMF:
> Anyway, build it failing with some new hell. You probably need to copy the `project_config` configuration from one of the other container builders in another extensions.

Okay. Will try tomorrow or something.

dantleech:
thanks!

PR #37, opened 2021-04-25
Rename handler

dantleech:
Implement the `workspace/willRename` handler to get the workspace edits necessary to move all the files.

NOTE:

- VSCode does not seem to support this at present.
- CoC will get support [soon](https://github.com/neoclide/coc.nvim/issues/3050)

This PR removes the "file/folder" rename detection introdued recently, as it does not work well with the index in it's current state.


dantleech:
cc @BladeMF

lib/LanguageServerRename/Handler/FileRenameHandler.php
```diff
@@ -0,0 +1,74 @@
...
+        ];
+    }
+
+    /**
+     * @return Promise<WorkspaceEdit>
+     */
+    public function willRenameFiles(RenameFilesParams $params): Promise
+    {
+        return call(function () use ($params) {
+            $workspaceEdits = LocatedTextEditsMap::create();
```
dantleech, line 55: note that we should now replace these VO's with the new WorkspaceEdits and TextDocumentEdit VOs from the text-document package.

PR #38, opened 2021-05-03
Generate Method code action

BladeMF:
A code action for the "Generate method" refactoring. Some points (also noted in the code):
* It doesn't provide diagnostics (although the code is there, the tag is commented out from the container setup), because it doesn't check method calls for "undefined method". I wanted to push v1 out and mess around with (potentially) slow calls to Worse reflection for v2. So the whenever the code for `provideDiagnostics` is fitted with "check for undefined method", it will be good to go;
* As a result, if a method is defined and the code action is executed, nothing happens;
* It shows up on all method calls;
* Otherwise it's really nice to use.

The test I intentionally made a **unit** test rather than an **integration** test, because:
* Integration tests are really complicated to setup and I rather not do this without a reason;
* A single integration case for a `CodeActionProvider` and `DiagnosticsProvider` will prove the plumbing is working - I'd rather test each `CodeActionProvider` whether they provide the correct actions/diagnostics in the correct context and then test whether the `Command` is passing to and from the model and handling exceptions.

The only problem with the above is the very difficult mocking of the `ClientApi` class. We can create a `ClientApiSpy` class that collects all calls to `public function request(string $method, array $params): Promise` and `public function notification(string $method, array $params): void` methods or allows a callback to emulate responses.

### DIFF
--- a/lib/LanguageServerCodeTransform/CodeAction/GenerateMethodProvider.php
+++ b/lib/LanguageServerCodeTransform/CodeAction/GenerateMethodProvider.php
@@ -3,21 +3,35 @@
 namespace Phpactor\Extension\LanguageServerCodeTransform\CodeAction;
 
 use Amp\Promise;
+use Amp\Success;
+use Microsoft\PhpParser\Node\Expression\CallExpression;
+use Microsoft\PhpParser\Node\Expression\MemberAccessExpression;
+use Microsoft\PhpParser\Parser;
+use Microsoft\PhpParser\Token;
 use Phpactor\Extension\LanguageServerBridge\Converter\PositionConverter;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\GenerateMethodCommand;
 use Phpactor\LanguageServerProtocol\CodeAction;
 use Phpactor\LanguageServerProtocol\Command;
+use Phpactor\LanguageServerProtocol\Diagnostic;
+use Phpactor\LanguageServerProtocol\DiagnosticSeverity;
+use Phpactor\LanguageServerProtocol\Position;
 use Phpactor\LanguageServerProtocol\Range;
 use Phpactor\LanguageServerProtocol\TextDocumentItem;
 use Phpactor\LanguageServer\Core\CodeAction\CodeActionProvider;
+use Phpactor\TextDocument\ByteOffset;
 use function Amp\call;
 
 class GenerateMethodProvider implements CodeActionProvider
 {
     public const KIND = 'quickfix.generate_method';
+    /**
+     * @var Parser
+     */
+    private $parser;
 
-    public function __construct()
+    public function __construct(Parser $parser)
     {
+        $this->parser = $parser;
     }
     /**
      * {@inheritDoc}
@@ -27,6 +41,13 @@ class GenerateMethodProvider implements CodeActionProvider
         return [
              self::KIND
          ];
+    }
+    /**
+     * {@inheritDoc}
+     */
+    public function provideDiagnostics(TextDocumentItem $textDocument): Promise
+    {
+        return new Success($this->getDiagnostics($textDocument));
     }
     /**
      * {@inheritDoc}
@@ -54,5 +75,62 @@ class GenerateMethodProvider implements CodeActionProvider
                 ])
             ];
         });
+    }
+    /**
+     * @return array<Diagnostic>
+     */
+    private function getDiagnostics(TextDocumentItem $textDocument): array
+    {
+        $node = $this->parser->parseSourceFile($textDocument->text);
+        $diagnostics = [];
+        foreach ($node->getDescendantNodes() as $node) {
+            if ((!$node instanceof CallExpression)) {
+                continue;
+            }
+            assert($node instanceof CallExpression);
+            if ((!$node->callableExpression instanceof MemberAccessExpression)) {
+                continue;
+            }
+            
+            $memberName = $node->callableExpression->memberName;
+            if(!($memberName instanceof Token))
+                continue;
+            assert($memberName instanceof Token);
+            
+            $diagnostics[] = new Diagnostic(
+                new Range(
+                    PositionConverter::byteOffsetToPosition(
+                        ByteOffset::fromInt($memberName->start),
+                        $textDocument->text
+                     ),
+                    PositionConverter::byteOffsetToPosition(
+                        ByteOffset::fromInt($memberName->start + $memberName->length), 
+                        $textDocument->text
+                    ),
+                ),
+                'Generate method',
+                DiagnosticSeverity::INFORMATION,
+                null,
+                'phpactor'
+            );
+        }
+
+        usort($diagnostics, function(Diagnostic $a, Diagnostic $b){
+            if($a->range->start->line > $b->range->start->line)
+                return 1;
+            
+            if($a->range->start->line < $b->range->start->line)
+                return -1;
+            
+            if($a->range->start->character > $b->range->start->character)
+                return 1;
+
+            if($a->range->start->character < $b->range->start->character)
+                return -1;
+            
+            return 0;
+        });
+
+        return $diagnostics;
     }
 }
--- a/lib/LanguageServerCodeTransform/CodeAction/GenerateMethodProvider.php
+++ b/lib/LanguageServerCodeTransform/CodeAction/GenerateMethodProvider.php
@@ -6,6 +6,7 @@ use Amp\Promise;
 use Amp\Success;
 use Microsoft\PhpParser\Node\Expression\CallExpression;
 use Microsoft\PhpParser\Node\Expression\MemberAccessExpression;
+use Microsoft\PhpParser\Node\Expression\ScopedPropertyAccessExpression;
 use Microsoft\PhpParser\Parser;
 use Microsoft\PhpParser\Token;
 use Phpactor\Extension\LanguageServerBridge\Converter\PositionConverter;
@@ -14,12 +15,13 @@ use Phpactor\LanguageServerProtocol\CodeAction;
 use Phpactor\LanguageServerProtocol\Command;
 use Phpactor\LanguageServerProtocol\Diagnostic;
 use Phpactor\LanguageServerProtocol\DiagnosticSeverity;
-use Phpactor\LanguageServerProtocol\Position;
 use Phpactor\LanguageServerProtocol\Range;
 use Phpactor\LanguageServerProtocol\TextDocumentItem;
 use Phpactor\LanguageServer\Core\CodeAction\CodeActionProvider;
+use Phpactor\LanguageServer\Core\Diagnostics\DiagnosticsProvider;
 use Phpactor\TextDocument\ByteOffset;
 use function Amp\call;
+use function array_filter;
 
 class GenerateMethodProvider implements DiagnosticsProvider, CodeActionProvider
 {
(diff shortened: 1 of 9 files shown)


lib/LanguageServerCodeTransform/CodeAction/GenerateMethodProvider.php
```diff
@@ -0,0 +1,168 @@
...
+                array_filter($diagnostics, function (Diagnostic $diag) use ($range) {
+                    return (
+                        $diag->range->start->line <= $range->start->line &&
+                        $diag->range->start->character <= $range->start->character &&
+                        $diag->range->end->line >= $range->end->line &&
+                        $diag->range->end->character >= $range->end->character
+                    );
+                })
+            );
+
```
BladeMF, line 80: The above probably needs to change to a code similar to what diagnostics does. I am generally fearful of performance similar to that of Import Names where it gets really slow. I wonder if we can cache the result of the diagnostics and use that to see whether we're on one? If the cache is not there we can always get them again.

lib/LanguageServerCodeTransform/CodeAction/GenerateMethodProvider.php
```diff
@@ -0,0 +1,168 @@
...
+            if (count($diagnostics) === 0) {
+                return [];
+            }
+
+            return [
+                CodeAction::fromArray([
+                    'title' =>  'Generate method (if not exists)',
```
BladeMF, line 87: This kind of goes with the fact that we're not checking the method calls.

lib/LanguageServerCodeTransform/CodeAction/GenerateMethodProvider.php
```diff
@@ -0,0 +1,168 @@
...
+                        ]
+                    )
+                ])
+            ];
+        });
+    }
+    /**
+     * @return array<Diagnostic>
+     */
+    private function getDiagnostics(TextDocumentItem $textDocument): array
```
BladeMF, line 105: Diagnostics are there, but they are not used in the LSP as they would underline every method call.

lib/LanguageServerCodeTransform/CodeAction/GenerateMethodProvider.php
```diff
@@ -0,0 +1,168 @@
...
+                    ),
+                ),
+                'Generate method',
+                DiagnosticSeverity::INFORMATION,
+                null,
+                'phpactor'
+            );
+        }
+
+        usort($diagnostics, function (Diagnostic $a, Diagnostic $b) {
```
BladeMF, line 146: If not sorted it becomes untestable because the order that `TolerantParser` returns the method calls is sometimes different from the order they occur in the source (purely as line/char).

lib/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
```diff
@@ -160,5 +176,15 @@ private function registerCodeActions(ContainerBuilder $container): void
...
             LanguageServerExtension::TAG_CODE_ACTION_PROVIDER => []
         ]);
+
+        $container->register(GenerateMethodProvider::class, function (Container $container) {
+            return new GenerateMethodProvider(
+                // $container->get(CodeTransformExtension::SERVICE_CLASS_GENERATORS),
+                $container->get('worse_reflection.tolerant_parser')
+            );
+        }, [
+            // LanguageServerExtension::TAG_DIAGNOSTICS_PROVIDER => [],
```
BladeMF, line 184: This is how the diagnostics are turned off.

lib/LanguageServerHover/Handler/HoverHandler.php
```diff
@@ -73,16 +73,19 @@ public function hover(
...
-                PositionConverter::byteOffsetToPosition(
-                    ByteOffset::fromInt($symbolContext->symbol()->position()->start()),
-                    $document->__toString()
-                ),
-                PositionConverter::byteOffsetToPosition(
-                    ByteOffset::fromInt($symbolContext->symbol()->position()->end()),
-                    $document->__toString()
-                )
+                $position,
+                $position
```
BladeMF: Not sure what to do about this one. It is really annoying but it probably belongs in its own PR.

tests/LanguageServerCodeTransform/Unit/LspCommand/GenerateMethodCommandTest.php
```diff
@@ -0,0 +1,173 @@
...
+        $sourceCode = SourceCode::fromStringAndPath(
+            $textDocumentItem->text,
+            $textDocumentUri->path()
+        );
+        $textEdits = new TextDocumentEdits(
+            $textDocumentUri,
+            new TextEdits(TextEdit::create(5, 1, 'test'))
+        );
+
+        $rpcClient = new class() implements RpcClient {
```
BladeMF: By doing a mock `RpcClient` we can intercept calls to `notification` and `request` and then spy on them on mock responses.

lib/LanguageServerCodeTransform/LspCommand/GenerateMethodCommand.php
```diff
@@ -61,6 +62,9 @@ public function __invoke(string $uri, int $offset): Promise
         } catch (MethodCallNotFound $error) {
             $this->clientApi->window()->showMessage()->warning($error->getMessage());
             return new Success(null);
+        } catch (CouldNotResolveNode $error) {
```
dantleech: i think both of these extend `NotFound` ?

lib/LanguageServerCodeTransform/CodeAction/GenerateMethodProvider.php
dantleech: Why not detect if the method exists or not?

lib/LanguageServerHover/Handler/HoverHandler.php
dantleech: yes if not related please remove

tests/LanguageServerCodeTransform/Unit/LspCommand/GenerateMethodCommandTest.php
dantleech: you can use the test language server builder for this I think

dantleech: https://github.com/phpactor/language-server-phpactor-extensions/blob/adf9f1857732cb0e1365d38f7614313ccf0f8f7a/tests/LanguageServerCodeTransform/Unit/CodeAction/CreateClassProviderTest.php#L65

the transmitter has all the messages that were transmitted from the LS

tests/LanguageServerCodeTransform/Unit/LspCommand/GenerateMethodCommandTest.php
```diff
@@ -0,0 +1,186 @@
...
+            }
+        };
+        $apiClient = new ClientApi($rpcClient);
+
+        $workspace = $this->prophesize(Workspace::class);
+        // @phpstan-ignore-next-line
+        $workspace->get($uri)
+            ->willReturn($textDocumentItem);
+
+        $generateMethod = $this->prophesize(GenerateMethod::class);
```
dantleech, line 51: if using the test language server you don't need to mock the workspace:

```
        $languageServer->textDocument()->open($url, $content)
```

lib/LanguageServerCodeTransform/LspCommand/GenerateMethodCommand.php
BladeMF: Indeed you are correct.

lib/LanguageServerCodeTransform/CodeAction/GenerateMethodProvider.php
BladeMF: a) Was worried about performance
b) Wanted to get the other part running
c) Didn't know what was the best way to do that.

If you can help with c), that'll be great.

tests/LanguageServerCodeTransform/Unit/LspCommand/GenerateMethodCommandTest.php
BladeMF: Did it using `TestRpcClient`.

BladeMF:
Not sure why 7.3 CS fixer fails...

dantleech:
restarted the build, probably temporary failure. you should be able to use the LS server tester more than you have, but it's already much better by using RpcTestClient.

dantleech:
:boom: I will see if I can iterate on this for diagnostics and the tests.

tests/LanguageServerCodeTransform/Unit/CodeAction/GenerateMethodProviderTest.php
```diff
@@ -0,0 +1,215 @@
...
+use Phpactor\LanguageServerProtocol\DiagnosticSeverity;
+use Phpactor\LanguageServerProtocol\Position;
+use Phpactor\LanguageServerProtocol\Range;
+use Phpactor\LanguageServerProtocol\TextDocumentItem;
+use Phpactor\TextDocument\ByteOffsetRange;
+use function Amp\Promise\wait;
+
+class GenerateMethodProviderTest extends TestCase
+{
+    /** @dataProvider provideDiagnosticsTestData */
```
dantleech, line 24: for future reference this should be:
```
/**
 * @dataProvider blah
 * /
 */
```

for consistence.

tests/LanguageServerCodeTransform/Unit/CodeAction/GenerateMethodProviderTest.php
```diff
@@ -0,0 +1,215 @@
...
+                        GenerateMethodCommand::NAME,
+                        [
+                            $uri,
+                            $result->offset('selection')->toInt()
+                        ]
+                    )
+                ])
+            ];
+        } else {
+            $codeActions = [];
```
dantleech, line 130: please try and avoid `else` statements in the future. `if (!$shouldSucceed) { self::assertEmpty($blah) }`

dantleech, line 130: You can avoid this IF/ELSE by delegating `$codeActions = $this->createCodeActions($shouldSucceed)` for example and returning early.

tests/LanguageServerCodeTransform/Unit/LspCommand/GenerateMethodCommandTest.php
```diff
@@ -0,0 +1,131 @@
...
+        self::assertNotNull($showMessage);
+        self::assertEquals([
+            'type' => 2,
+            'message' => $exception->getMessage()
+        ], $showMessage->params);
+    }
+
+    public function provideExceptions(): array
+    {
+        return [
```
dantleech, line 125: to be consistent ould be better in future to use `yield`

tests/LanguageServerCodeTransform/Unit/CodeAction/GenerateMethodProviderTest.php
BladeMF: I keep forgetting you want them multilined, sorry.

dantleech: it's just consistency, I should update the CS fixer to catch it.

PR #40, opened 2021-05-08
Generate method

dantleech:
This PR updates the method generator to use the ne `MissingMethodFinder` and act on the entire document:

- Invoking the code action anywhere offers the possiblity to geneate missing methods
- Diagnostics are enabled.

@BladeMF note that as we no longer use the parser, the tests have been simplified, the deleted test cases could be moved to code-transform however.

I also notice that, at least for me, generating methods in a foreign class does not place the method in the correct place (probably the source or offset is being wrongly mapped).

### DIFF
--- a/tests/LanguageServerCodeTransform/Unit/CodeAction/GenerateMethodProviderTest.php
+++ b/tests/LanguageServerCodeTransform/Unit/CodeAction/GenerateMethodProviderTest.php
@@ -3,24 +3,17 @@
 namespace Phpactor\Extension\LanguageServerCodeTransform\Tests\Unit\CodeAction;
 
 use Generator;
-use Microsoft\PhpParser\Parser;
 use PHPUnit\Framework\TestCase;
 use Phpactor\CodeTransform\Domain\Helper\MissingMethodFinder;
 use Phpactor\CodeTransform\Domain\Helper\MissingMethodFinder\MissingMethod;
-use Phpactor\Extension\LanguageServerBridge\Converter\PositionConverter;
-use Phpactor\Extension\LanguageServerBridge\Converter\RangeConverter;
 use Phpactor\Extension\LanguageServerCodeTransform\CodeAction\GenerateMethodProvider;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\GenerateMethodCommand;
-use Phpactor\Extension\LanguageServerRename\Tests\Util\OffsetExtractor;
 use Phpactor\LanguageServerProtocol\CodeAction;
 use Phpactor\LanguageServerProtocol\Command;
 use Phpactor\LanguageServerProtocol\Diagnostic;
 use Phpactor\LanguageServerProtocol\DiagnosticSeverity;
-use Phpactor\LanguageServerProtocol\Position;
-use Phpactor\LanguageServerProtocol\Range;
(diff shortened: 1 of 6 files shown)


tests/LanguageServerCodeTransform/Unit/CodeAction/GenerateMethodProviderTest.php
```diff
@@ -3,215 +3,133 @@
...
         self::assertEquals(
-            [],
-            wait($provider->provideActionsFor(new TextDocumentItem('file:///somefile.php', 'php', 1, '<?php $var = 5;'), new Range(
-                new Position(1, 7),
-                new Position(1, 9),
-            )))
+            $expectedActions,
+            wait($provider->provideActionsFor(
+                new TextDocumentItem(self::EXAMPLE_FILE, 'php', 1, self::EXAMPLE_SOURCE),
+                ProtocolFactory::range(0, 0, 0, 0)
```
dantleech, line 90: note that the range no longer matters

dantleech:
I've added the additional test cases to code-transform

BladeMF:
I have become really unsure of having diagnostics enabled. I mean, `phpstan` will tell me about undefined methods anyway, I really don't want blue or yellow or red squiggly lines under methods that phpactor can't resolve. I still support showing the action only on unresolvable methods though. 

What I suggest is at least make it an option to turn those off. I, as a developer, am ***really bugged*** by squggly lines - `phpstan`'s underscores on `MethodProphecy` methods drive me **NUTS**. What do you think?

dantleech:
> What I suggest is at least make it an option to turn those off

You can turn them off: https://phpactor.readthedocs.io/en/master/reference/configuration.html#language-server-diagnostics-on-update but I guess you mean more granularly?

BladeMF:
Yes, just these diagnostics - for `generate method`. Do you have diagnostics where you work? VIM? Having all unknown methods underlined is basically what `phpstan` does after a certain level.

BladeMF:
I guess I can try it and see.

My whole point is that this can be annoying if there are a lot of objects whose types are not inferred correctly (or, all prophecies, for that matter) - your whole file could light up with no way to turn it off, but those would not be generatable methods (or those would not light up?).

I think just having the light bulb (code actions) showing up after you step on an unknown method will be enough. This is how Visual Studio (not Code) does all refactorings - it offers them after you've selected the code or stepped on the property or method. So nobody would expect `phpactor` to highlight all possible generate method sites in the file in advance (which is diagnostics).

dantleech:
Yes we can proably remove the diagnostics or make them optional :)

BladeMF:
Prophecised methods get underscored :-( I actually don't mind the diagnostics elsewhere for now, but here it is annoying.
image
1. Should that be solved with generics? 
2. Probably the same mechanism that will give us the completion of the methods for `ObjectProphecy` objects?
3. Should we try to turn of diagnostics for those kind of objects for now?

dantleech:
I don't mind adding some mechanism to specific diagnostics. Generics AND an extension would solve this, but generics are still a long way away (I would need weeks of dedicated time probably)

BladeMF:
Not sure what to do either. Probably a "turn off diagnostics for undefined methods" setting would be the easiest solution.

PR #41, opened 2021-05-08
Extract method LSP code action

BladeMF:


lib/LanguageServerCodeTransform/LspCommand/ExtractMethodCommand.php
```diff
@@ -0,0 +1,67 @@
...
+    public function __invoke(string $uri, int $startOffset, int $endOffset): Promise
+    {
+        $textDocument = $this->workspace->get($uri);
+        
+        try {
+            $textEdits = $this->extractMethod->extractMethod(
+                SourceCode::fromStringAndPath($textDocument->text, $textDocument->uri),
+                $startOffset,
+                $endOffset,
+                self::DEFAULT_METHOD_NAME
```
dantleech, line 56: it seems we canot currently ask the user [for input](https://github.com/microsoft/language-server-protocol/issues/1164). Assuming this method will throw an exception if there is an existing method, you could add an increment to the default method name.

dantleech (approved):

lib/LanguageServerCodeTransform/LspCommand/ExtractMethodCommand.php
BladeMF: Actually my guess was, since `myMethod` is so silly of a name, if you have one of these lying around, you have some more work to do until you extract another one. Go rename the other one, _then_ extract this one :-) Kind of a code quality error. Otherwise doing a counter would be trivial, I think, if "method exists" is a catchable exception.

The perfect solution would be to either ask for input (I saw you found the same discussion), or enter rename mode immediately after completing. Probably it's best to wait for the former.

So, code quality wise, I'd leave it as it is, but I am happy to do a counter if think it'll be better.

PR #45, opened 2021-05-23
Extract expression

BladeMF:
Classic stuff.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "dms/phpunit-arraysubset-asserts": "dev-master",
         "ergebnis/composer-normalize": "^2.0",
         "friendsofphp/php-cs-fixer": "^2.17",
-        "jangregor/phpstan-prophecy": "^0.8.1",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -39,7 +39,7 @@
         "dms/phpunit-arraysubset-asserts": "dev-master",
         "ergebnis/composer-normalize": "^2.0",
         "friendsofphp/php-cs-fixer": "^2.17",
-        "jangregor/phpstan-prophecy": "^0.8.0",
+        "jangregor/phpstan-prophecy": "^0.8.1",
```
dantleech: Is there a reason to bump this? `^` will automatically use `8.1`

BladeMF: No. For some reason it was really hard to install it or something. It was not running so I thought it wasn't there. So I did `composer require` and it thought I wanted `update`. So no, I will bring it back. The only change that needs to happen is in `phpstan.neon`.

BladeMF: Interesting - `phpactor` uses 0.8.1.

PR #46, opened 2021-05-23
Import all names

dantleech:
Additional code action from `ImportName`  provider to trigger new Command `ImportAllUnresolvedNames`.

- Will automatically add all unresolved names with exactly 1 candidate
- Will prompt user to resolve ambiguous imports

image


BladeMF:
I really need to track down these multiple suggestions in the import menu.

PR #48, opened 2021-05-24
Prevent multiple candidates for the same unresolved name

### LINKED ISSUE #1253: Suggestions for importing classes show items multiple times
image

Also, it is sometimes extremely slow.

BladeMF:
Fixes https://github.com/phpactor/phpactor/issues/1253.

### DIFF
--- a/lib/LanguageServerCodeTransform/CodeAction/ImportNameProvider.php
+++ b/lib/LanguageServerCodeTransform/CodeAction/ImportNameProvider.php
@@ -144,6 +144,7 @@ class ImportNameProvider implements CodeActionProvider, DiagnosticsProvider
 
     private function codeActionForFqn(NameWithByteOffset $unresolvedName, string $fqn, TextDocumentItem $item): CodeAction
     {
+        $diagnostics = $this->diagnosticsFromUnresolvedName($unresolvedName, $item, true);
         return CodeAction::fromArray([
             'title' => sprintf(
                 'Import %s "%s"',
@@ -152,7 +153,7 @@ class ImportNameProvider implements CodeActionProvider, DiagnosticsProvider
             ),
             'kind' => 'quickfix.import_class',
             'isPreferred' => false,
-            'diagnostics' => $this->diagnosticsFromUnresolvedName($unresolvedName, $item, true),
+            'diagnostics' => ($diagnostics[1] !== null) ? [$diagnostics[1]] : null,
             'command' => new Command(
                 'Import name',
                 ImportNameCommand::NAME,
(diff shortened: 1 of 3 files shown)


lib/LanguageServerCodeTransform/Model/NameImport/CandidateFinder.php
```diff
@@ -66,12 +66,16 @@ public function unresolved(TextDocumentItem $item): NameWithByteOffsets
...
         TextDocumentItem $item
     ): Generator {
-        $actions = [];
+        $added = [];
         foreach ($this->unresolved($item) as $unresolvedName) {
+            assert($unresolvedName instanceof NameWithByteOffset);
+            $nameString = (string)$unresolvedName->name();
+            if (array_search($nameString, $added, true) !== false) {
+                continue;
+            }
```
dantleech, line 75: You can use a dict instead, which avoids searching the array:

```
$seen = [];
if (isset($seen[$hash])) {
    continue;
}

$seen[$hash] = true;
```

If the byte offest is not required then we can just return an array of strings....

BladeMF: My reasoning is that byte offset is important for diagnostics only. That's why for one code action there can be multiple diagnostics (not sure we're doing that btw).

BladeMF: Found one more optimisation, BTW. Wait for it.

BladeMF: Done:
1. When providing diagnostics, there were candidate searches for each diagnostic. So cache the fact that there were or were not candidates for each name. Reuse it for next diagnostic for the same name and another offset.
2. When providing code actions, there were needless searches for candidates to display the diagnostic as a hint although we already _know_ there _are_ candidates. So use the same mechanism to not search but assume there are candidates and  the diagnostics were hints.

BladeMF: Currently we're just not assigning all diagnostics that relate to a code action, which I view as a minor problem.

lib/LanguageServerCodeTransform/CodeAction/ImportNameProvider.php
```diff
@@ -87,10 +96,13 @@ private function diagnosticsFromUnresolvedName(NameWithByteOffset $unresolvedNam
             )
         );
 
-        $candidates = iterator_to_array($this->finder->candidatesForUnresolvedName($unresolvedName));
+        if ($hasCandidates === null) {
+            $hasCandidates = $this->finder->candidatesForUnresolvedName($unresolvedName)->current() !== null;
+        }
 
-        if (count($candidates) === 0) {
+        if ($hasCandidates) {
```
dantleech: Should this be `false === $hasCandidates`? If it has candidates then it can be imported, this expression leads to "clss doesn't exist"

BladeMF: It would seem that I really did a half-arsed job. You are correct.

BladeMF:
Also ping here, @dantleech .

BladeMF:
Ping, @dantleech. Hope everything is okay with you.

dantleech:
Sorry, haven't had much capacity and have been on holiday will test the latest changes when I get internet that allows SSH

PR #50, opened 2021-06-19
Replace completion command for imports by additionalTextEdits

weeman1337:
@dantleech maybe you could put this into your local setup and test it additionally to me since this is a critical place. Completions and imports are nearly used every day (even every minute :D - at least by me).

I did my very best for testing
- local editor setup
- test cases extended / adapted

With these changes phpactor/phpactor#1300 doesn't appear any more.

### DIFF
--- a/lib/LanguageServerCodeTransform/Model/NameImport/NameImporter.php
+++ b/lib/LanguageServerCodeTransform/Model/NameImport/NameImporter.php
@@ -34,6 +34,7 @@ class NameImporter implements Command
         int $offset,
         string $type,
         string $fqn,
+        bool $updateReferences,
         ?string $alias = null
     ): NameImporterResult {
         $sourceCode = SourceCode::fromStringAndPath(
@@ -46,16 +47,17 @@ class NameImporter implements Command
             NameImport::forClass($fqn, $alias);
 
         try {
-            $textEdits = $this->importName->importName(
-                $sourceCode,
-                ByteOffset::fromInt($offset),
-                $nameImport
-            );
+            $byteOffset = ByteOffset::fromInt($offset);
+            if ($updateReferences) {
+                $textEdits = $this->importName->importName($sourceCode, $byteOffset, $nameImport);
+            } else {
+                $textEdits = $this->importName->importNameOnly($sourceCode, $byteOffset, $nameImport);
+            }
             $lspTextEdits = TextEditConverter::toLspTextEdits($textEdits, $document->text);
             return NameImporterResult::createResult($nameImport, $lspTextEdits);
         } catch (NameAlreadyImportedException $error) {
-            if ($error->existingName() === $fqn) {
-                return NameImporterResult::createEmptyResult();
+            if ($error->existingFQN() === $fqn) {
+                return $this->createResultForAlreadyImportedFQN($nameImport, $error);
             }
 
             $name = FullyQualifiedName::fromString($fqn);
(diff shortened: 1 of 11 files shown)


lib/LanguageServerCodeTransform/Model/NameImport/NameImporter.php
```diff
@@ -46,16 +47,17 @@ public function __invoke(
...
-                $sourceCode,
-                ByteOffset::fromInt($offset),
-                $nameImport
-            );
+            $byteOffset = ByteOffset::fromInt($offset);
+            if ($updateReferences) {
+                $textEdits = $this->importName->importName($sourceCode, $byteOffset, $nameImport);
+            } else {
+                $textEdits = $this->importName->importNameOnly($sourceCode, $byteOffset, $nameImport);
+            }
```
dantleech: hate to be pedantic: but can we remove the `else` and have one assignment?  `$textEdits = $this->importNameTextEdits($a, $b, $c)`?

lib/LanguageServerCodeTransform/Model/NameImport/NameImporter.php
```diff
@@ -64,12 +66,29 @@ public function __invoke(
...
     }
+
+    private function createResultForAlreadyImportedFQN(
+        NameImport $nameImport,
+        NameAlreadyImportedException $error
+    ): NameImporterResult {
+        if ($error->existingName() !== $nameImport->name()->head()->__toString()) {
+            $alias = $error->existingName();
+        } else {
+            $alias = null;
```
dantleech: could remove the else and iniitalize `$alias` as `null`

lib/LanguageServerCompletion/Handler/CompletionHandler.php
```diff
@@ -100,29 +109,42 @@ public function completion(CompletionParams $params, CancellationToken $token):
             $items = [];
             $isIncomplete = false;
             foreach ($suggestions as $suggestion) {
+                /** @var Suggestion $suggestion */
```
dantleech: typically using `assert($suggestion instanceof Suggestion` here. It helps with refactoring.

lib/LanguageServerCompletion/Handler/CompletionHandler.php
```diff
@@ -100,29 +109,42 @@ public function completion(CompletionParams $params, CancellationToken $token):
...
+
+                $parentNode = $this->parser
+                        ->parseSourceFile($textDocument->text)
+                        ->getDescendantNodeAtPosition($byteOffset->toInt())
+                        ->getParent();
```
dantleech: hmm, ideally we should not be parsing anything here. can this logic be in code-transform?

dantleech:
Confirm that this fixes the issue I reported earlier :+1:

lib/LanguageServerCompletion/Handler/CompletionHandler.php
weeman1337: Then code-transform needs to report back the "context". E.g. `CONTEXT_USE`, `CONTEXT_OTHER` or something like that. Completion has to act depending on the context. Or do you have another idea?

dantleech: Not sure I understand the `if` condtiion: if we are in a namespace declaration block, then don't add teh text edit? But why then would the completion return a class import?

weeman1337: > if we are in a namespace declaration block, then don't add teh text edit?

Yes. An import doesn't make sense here.

>  But why then would the completion return a class import?

Happens here: https://github.com/phpactor/completion/blob/master/lib/Core/Completor/NameSearcherCompletor.php#L82 . Do you think this is the place to check for the namespace declaration block?

dantleech: Rather here? https://github.com/phpactor/completion/blob/master/lib/Core/Completor/NameSearcherCompletor.php#L58

i.e. we have the/a node here.

weeman1337: I removed the parsing from the `CompletionHandler` and changed the `NameSearcherCompletor` to not return imports in `use` context, see phpactor/completion#58 .

lib/LanguageServerCodeTransform/Model/NameImport/NameImporter.php
weeman1337: Done

lib/LanguageServerCodeTransform/Model/NameImport/NameImporter.php
weeman1337: Done

lib/LanguageServerCompletion/Handler/CompletionHandler.php
weeman1337: Done

dantleech:
Needs `php-cs-fixer fix`

weeman1337:
> Needs php-cs-fixer fix

Puh - I did that; but with PHP 7.4.

Why does running the cs-fixer reports braces to be set different depending on using PHP 7.3 / PHP 7.4? :/

dantleech:
:shrug:

weeman1337:
@dantleech I don't want to stress you :D Last time you mentioned you have forgotten a PR. So see this as a friendly reminder.

I am desperately waiting for this one. After that I can switch to Phpactor.

dantleech:
Thanks, forgotten indeed....

dantleech:
I've updated phpactor master now too

PR #51, opened 2021-09-19
IndexerHandler: use ProgressNotifier

camilledejoye:
Depends on https://github.com/phpactor/language-server-extension/pull/23

Make use of the new `ProgressNotifier` in the `IndexHandler`.

@dantleech I introduced a BC break by adding the new dependency.
Do you prefer that I handle it gracefully and trigger a depreciation if the service is not provided ?

It works on my side, the only annoying thing is that the last message disappear right away but ti's a client configuration issue IMO and it should not prevent us from using this feature on our side.


dantleech:
It's fine to break BC here this isn't a reusable package

dantleech (approved):

PR #54, opened 2022-01-21
Fix checking missing didChangeWatchedFiles capability

przepompownia:
It only fixes casting to bool to avoid `PHP Warning:  Undefined array key "didChangeWatchedFiles"` in logs. Neovim's LSP built-in client [does not send it](https://github.com/neovim/neovim/issues/16078).


dantleech (approved):

dantleech:
Thanks!

PR #55, opened 2022-01-21
Use TextDocumentUri to create SymbolInformation

przepompownia:
It allows to use responses from `workspace/symbol` by nvim built-in LSP client. See https://github.com/neovim/neovim/blob/70db972e5fbcab39946ad8ac05472a693cf65b68/runtime/lua/vim/uri.lua#L105


dantleech:
🤞

